PR #30, opened 2020-08-05
Update to Sequelize 6.x and fixed a admin-ui/setting related bug

luxn:
Pull-Reuquest to update the outdated sequelize version to the current one. Sequelize's breaking API changes were treated.

Background:
I simply checked out the repo, but the migrations won't run for me. Looks like I have used the latest nodejs (I'm using node v14) where Sequelize/cli got problems with the dependencies inside the project. https://github.com/sequelize/sequelize/issues/12158
So I updated them.

By trying out the admin ui, I found that the Default Settings were not saved when I was editing them. Found out that there is no kinda upsert and if the setting value was not in the database table, the setting would not be updating. There is also no error indicator present. so if Settings were not pre inserted into the database the changes in admin ui have no effect, because they are only inserts (and they were returning 0 affected rows in the response). The added code ensures that the default settings are present in the database.


dorshay6:
Overall it looks good, I see some issues with eslint on that branch, I will add GitHub action to run lint in here

server/lib/settings/index.js
```diff
@@ -41,14 +41,48 @@ const getAllSettingFromDb = async () => {
   return settingsList;
 };
 
+function getSettingValueType(settingValue) {
```
oriefrati: please use arrow function
const getSettingValueType = (settingValue)  => {
...
}

server/migrations/20200223102246-add-deleted-at-index.js
```diff
@@ -1,17 +1,23 @@
...
-    await queryInterface.addIndex('users', 'users_email');
-    await queryInterface.addConstraint('users', 'phone_number');
+    await queryInterface.removeIndex('users',
+      ['phone_number', 'deleted_at'], {
+        type: 'UNIQUE'
+      });
+    await queryInterface.removeIndex('users',
+      ['email', 'deleted_at'], {
+        type: 'UNIQUE'
+      });
```
oriefrati, line 21: we should not change deployed migrations, please create new migration file and add your changes

server/lib/settings/index.js
```diff
@@ -41,14 +41,48 @@ const getAllSettingFromDb = async () => {
...
+    Object.keys(settingsList).map((defaultSettingKey) => {
+      if (settings.map((s) => s.key).includes(defaultSettingKey)) {
+        settingsList[defaultSettingKey] = settings.filter(s => s.key === defaultSettingKey)[0].value;
+      } else {
+        const defaultSettingToStore = Setting.build({
+            key: defaultSettingKey,
+            value: settingsList[defaultSettingKey].toString(),
+            type: getSettingValueType(settingsList[defaultSettingKey])
+        });
+        settingsToInsert.push(defaultSettingToStore.save());
```
oriefrati: you can create array of objects and use blukCreate instead of promise.all

oriefrati (changes requested):

server/migrations/20200223102246-add-deleted-at-index.js
luxn: I cannot add a new migration file because I only fixed the existing migration file here because the Sequelize API has changed. The old `await queryInterface.addConstraint('users', ['email', 'deleted_at'], { type: 'UNIQUE' })` does not accept the fields parameter (['email', 'deleted_at']). It was moved to the options parameter. This API Call ist now:
`await queryInterface.addConstraint('users', { fields: ['email', 'deleted_at'],  type: 'UNIQUE' })` Without the Change it was not possible (for me) to migrate up from base(empty db) with the new sequelize version.

See v6 https://sequelize.org/master/class/lib/dialects/abstract/query-interface.js~QueryInterface.html#instance-method-addConstraint 
And v4 https://sequelize.org/v4/class/lib/query-interface.js~QueryInterface.html#instance-method-addConstraint
The parameter is not existing anymore.

Secondly the call to `await queryInterface.removeIndex('users', 'users_email');` and `await queryInterface.sequelize.query('ALTER TABLE users DROP CONSTRAINT users_phone_number_key;');` is useless because this index / constraint is never created in a migration relative to this one before. The equivalent in the down function, why to add an index if it wasn't removed because it was not even there before.

server/lib/settings/index.js
luxn: done

server/lib/settings/index.js
luxn: done

OfryL:
Closed due inactivity

PR #39, opened 2020-09-05
fix server & disable stations

OfryL:


server/lib/super-express.js
```diff
@@ -13,7 +13,7 @@ const defaultOptions = {
   helmet: true,
 };
 
-module.exports = (options = {}) => {
+module.exports = function (options = {}) {
```
oriefrati, line 16: why not arrow function?

oriefrati (approved):

server/lib/super-express.js
OfryL: server/app.js:8 we construct it, so arrow func not working.

dorshay6 (approved):

PR #47, opened 2020-10-26
fix react zero bug

OfryL:


OfryL:
FYI the reason -
image

PR #106, opened 2022-05-31
refactor to ride context

OfryL:
## ℹ️ Change description:
Nothing here


locomotion-sdk/src/services/geo.js
```diff
@@ -95,5 +96,21 @@ class Geo {
...
 
-export default new Geo();
+const GeoService = new Geo();
+export default GeoService;
+
 export const { decodeGmPath } = Geo;
+
+export const getPosition = async () => {
+  try {
+    return GeoService.currentLocation();
```
itayankri, line 106: try/catch is useless here unless you replace the return with await and return the result after the try/catch.

itayankri (approved):

locomotion-sdk/src/services/geo.js
OfryL: agreed, but just a moving refactor.

PR #136, opened 2022-06-15
move turf from dev dep to prod dep to fix build issue

OmerGery:


.nvmrc
```diff
@@ -1 +1 @@
-14
+16
```
OfryL, line 1: you sure?

OfryL (approved):

.nvmrc
OmerGery: yes because npm i  on 14 moves package-lock to version 1 and does a lot of dep changes in package-lock , npm i on 16 doesn't change anything

PR #242, opened 2022-07-06
add cancel ride;

lieblichaaron:
## ℹ️ Change description:
Nothing here

<p>
  

</p>
</details>

Screen Shot 2022-07-07 at 10 51 33


examples/client/Locomotion/src/pages/Payments/index.js
```diff
@@ -45,14 +44,14 @@ export default ({ navigation, menuSide }) => {
     await usePayments.detachPaymentMethod(methodForDelete);
     await usePayments.loadCustomer();
     setLoading(false);
-    togglePopup('removeCard', false);
+    setIsCancelPopupVisible(false);
```
avivbarniv: Why?

lieblichaaron: because that is an outdated way we used to handle popups, its the only popup in that app still working through that context. either we should have them all that way or none

avivbarniv: So change the naming, it's not a cancel popup

examples/client/Locomotion/src/pages/Payments/index.js
```diff
@@ -96,14 +95,15 @@ export default ({ navigation, menuSide }) => {
           </CardContainer>
         )}
         <ConfirmationPopup
-          name="removeCard"
```
shirmagen, line 95: what about togglePopup('removeCard', true);? 
should be setIsCancelPopupVisible(true)?

examples/client/Locomotion/src/pages/Payments/index.js
lieblichaaron: my bad, fixed

PR #373, opened 2022-07-20
Aut 8986

lieblichaaron:
## ℹ️ Change description:
Nothing here

<p>
  
</p>
</details>

[AUT-9303](https://autofleet.atlassian.net/browse/AUT-9303)


examples/client/Locomotion/src/pages/RideHistory/RideCard/index.js
```diff
@@ -74,41 +106,60 @@ export const RideListView = ({
...
-        {ride.driver && ride.state === RIDE_STATES.COMPLETED && (
-        <DriverCard
-          activeRide={false}
-          ride={ride}
-        />
-        )}
-      </DriverCardContainer>
-    </RideViewContainer>
+    <>
+      {tip || tip === undefined ? (
```
sephina, line 124: can tip be null? is it ok

sephina (approved):

examples/client/Locomotion/src/pages/RideHistory/RideCard/index.js
lieblichaaron: i only want it to load once its checked the tip, if there isn't it will be undefined so i put either it has a value or undefined. the starting state is null so i dont want it to load while null. any better way to do it?

PR #432, opened 2022-07-27
fix logo

shirmagen:


examples/client/Locomotion/src/pages/ContactUs/index.js
```diff
@@ -82,7 +82,7 @@ export default ({ navigation, menuSide }) => {
             />
             <ScrollView>
               <ContactUsPageLogoContainer style={Platform.OS === 'android' ? { shadowColor: '#000' } : {}}>
-                <ContactUsLogo resizeMode="cover" source={logo} />
+                <ContactUsLogo resizeMode="cetner" source={logo} />
```
avivbarniv: center?*

avivbarniv (commented):
You can set max-height/width.
Please check it with few icons from the bucket and add screenshots

avivbarniv (approved):

PR #436, opened 2022-07-27
fix map marker shadow

shirmagen:


examples/client/Locomotion/src/Components/Marker/styled.js
```diff
@@ -56,18 +56,21 @@ export const MarkerToolTipText = styled.Text`
...
    flex-direction: column;
    background-color: #f0f0f0;
    border-radius: 8px;
    padding: 3px;
    align-items: center;
-  shadow-offset: 0px 0px;
-  shadow-color: #04214f;
-  shadow-opacity: 0.4;
-  elevation: 4;
+   ${Platform.OS === 'android' && 'border-color: #bcbcbc'};
```
OfryL: move `;` inside

OfryL (changes requested):

examples/client/Locomotion/src/Components/Marker/styled.js
OfryL: dont use `&&` you'll get a false as a string

OmerGery (approved):

avivbarniv (approved):

OfryL (approved):

PR #598, opened 2022-11-14
Pickup time window

sephina:
## ℹ️ Change description:
Nothing here

<p>
  
</p>
</details>

### DIFF
--- a/examples/client/Locomotion/src/Components/BsPages/index.tsx
+++ b/examples/client/Locomotion/src/Components/BsPages/index.tsx
@@ -283,7 +283,7 @@ export const ConfirmPickupTime = (props: any) => {
   const date = moment(unconfirmedPickupTime).format('ddd, MMM Do');
   const isDateToday = moment(unconfirmedPickupTime).isSame(moment(), 'day');
   const afterTime = moment(unconfirmedPickupTime).format('h:mm A');
-  const windowSize = chosenService && chosenService.pickupWindowSizeInMinutes;
+  const windowSize = chosenService?.pickupWindowSizeInMinutes;
   const beforeTime = (windowSize && moment(unconfirmedPickupTime).add(windowSize, 'minutes').format('h:mm A')) || i18n.t('general.noTimeWindow');
   const startDate = moment(unconfirmedPickupTime).add(unconfirmedPickupTime ? 0 : minMinutesBeforeFutureRide, 'minutes').toDate();
   const [tempSelectedDate, setTempSelectedDate] = useState(startDate);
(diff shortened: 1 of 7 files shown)


examples/client/Locomotion/src/pages/ActiveRide/RideDrawer/RideOptions/RideButtons/styled.ts
```diff
@@ -24,7 +24,6 @@ justify-content: space-between;
 export const ButtonContainer = styled(Button).attrs({
   noBackground: true,
 })`
-    background-color: ${({ highlight }) => (highlight ? '#d3eefc' : 'transparent')};
     flex-direction: row;
```
avivbarniv, line 27: It might be needed, did you check all the uses for this button?

examples/client/Locomotion/src/Components/BsPages/index.tsx
```diff
@@ -344,13 +349,15 @@ export const ConfirmPickupTime = (props: any) => {
       </RoundedButton>
       <DatePickerPoppup
         testID="datePicker"
-        textColor={getTextColorForTheme()}
+        textColor="black"
```
avivbarniv, line 352: should it be hardcoded? what about dark mode?

examples/client/Locomotion/src/pages/ActiveRide/RideDrawer/RideOptions/RideButtons/styled.ts
sephina: I've added it but not using it anymore

examples/client/Locomotion/src/Components/BsPages/index.tsx
sephina: we don't have different colors for all new ride process when on dark mode

examples/client/Locomotion/src/context/newRideContext/index.tsx
```diff
@@ -370,6 +374,7 @@ const RidePageContextProvider = ({ children }: {
       const formattedEstimations = formatEstimations(services, estimations, tags);
       setChosenService(ride.scheduledTo ? formattedEstimations.find((e: any) => e.currency)
         : formattedEstimations.find((e: any) => e.eta));
+      setDefaultService(formattedEstimations?.[0]);
```
avivbarniv, line 377: default service should be the same as chosen service

sephina: this is a case where are no available services, but we still want to use a service for a future ride. so there is no chosen service

avivbarniv (approved):

PR #602, opened 2022-11-14
remove edit button

noatzabar3:


examples/client/Locomotion/src/Components/BsPages/ActiveRide/share.tsx
```diff
@@ -34,7 +34,6 @@ const ShareButton = () => {
       onPress={() => {
         onShare();
       }}
-      style={{ width: ride.cancelable ? HALF_WIDTH : '100%' }}
```
itayankri, line 37: why did you remove it? this button will be 100% width all the time?

PR #630, opened 2022-12-19
add language picker. add greek and fake french

sephina:
## ℹ️ Change description:
Nothing here

<p>
  
</p>
</details>


examples/client/Locomotion/src/I18n/index.js
```diff
@@ -1,16 +1,65 @@
...
 import en from './en.json';
+import fr from './fr.json';
+import elGR from './el_GR.json';
+
+
+const USER_LANGUAGE_STORAGE_KEY = 'userLanguage';
+
+const getUserLanguage = async () => {
+  const languageCode = await StorageService.get(USER_LANGUAGE_STORAGE_KEY);
+  console.log('~~~ getUserLanguage', languageCode);
```
OmerGery: needed?

examples/client/Locomotion/src/popups/Selector/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+          <CloseButton onPress={async () => {
+            setSelectedItem(selected);
+            onCancel();
+          }}
+          />
+        </TitleView>
+        <ItemsScrollView>
+          <Container>
+            {items.map((item: any, i) => (
+              <Item
```
OmerGery, line 78: testID?

OmerGery (approved):

examples/client/Locomotion/src/popups/Selector/index.tsx
sephina: part of the button component (that item implements)

examples/client/Locomotion/src/I18n/index.js
```diff
@@ -1,22 +1,142 @@
...
-      ns: ['translation'],
-      defaultNS: 'translation',
-      interpolation: {
-        escapeValue: false, // not needed for react
+    backend: {
+      localResources,
+      remoteBackend: {
+        type: i18nHttpLoader,
+        options: {
+          loadPath: `https://storage.googleapis.com/language-files/locomotion/{{lng}}.json?timestamp=${moment().format('HH:mm')}`,
```
avivbarniv: Should be env var, it's a public code

examples/client/Locomotion/src/popups/Selector/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+                selected={selectedItem === i}
+                item={item}
+                onPress={() => {
+                  setSelectedItem(i);
```
avivbarniv, line 82: maybe use and id instead of index?

avivbarniv (approved):

examples/client/Locomotion/src/popups/Selector/index.tsx
sephina: but then to get the selected item we need to find it in the array instead of just get it. why do you think it's better to use the id?

PR #634, opened 2022-12-20
Version 3.2.17

OmerGery:


examples/client/Locomotion/.env.sample
```diff
@@ -10,3 +10,4 @@ DEFAULT_LONGITUDE=
 STRIPE_PUBLISHER_KEY=
 DONT_USE_STATIONS=
 OPERATION_ID=
+LANGUAGE_FILES_STORAGE=
```
oriefrati, line 13: the bucket is public?

examples/client/Locomotion/src/I18n/index.js
```diff
@@ -1,22 +1,146 @@
...
+
+const languageDetector = {
+  type: 'languageDetector',
+  async: true, // async detection
+  // eslint-disable-next-line no-return-await
+  detect: async cb => cb(await getPreferredLanguageCode()),
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  init: () => {},
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  cacheUserLanguage: () => {},
```
oriefrati, line 81: do we have to define it? maybe we can to delete instead of disable linter

examples/client/Locomotion/src/I18n/index.js
```diff
@@ -1,22 +1,146 @@
...
+export const getPreferredLanguageCode = async () => (
+  (await getUserLanguage())
+  || (Object.keys(supportedLanguages).includes(RNLocalize.getLocales()[0].languageCode)
+    ? RNLocalize.getLocales()[0].languageCode
+    : DEFAULT_LANGUAGE_CODE)
+);
+
+(async () => {
+  userLanguage = await getPreferredLanguageCode();
+})();
```
oriefrati: bad practice
we need to create function next version

oriefrati (approved):

examples/client/Locomotion/.env.sample
sephina: yes

examples/client/Locomotion/src/I18n/index.js
sephina: crashing w/o it

PR #640, opened 2022-12-29
Bug: Percentage should be after the number and not before

OmerGery:
before:
image

after:
image


sephina (commented):
why the name itself is not in the right order? what if we get coupon with correct order?

OmerGery:
> why the name itself is not in the right order? what if we get coupon with correct order?

yep , im checking the stripe api  , maybe i'll fix from backend

PR #655, opened 2023-01-10
fix item text

avivbarniv:
https://autofleet.atlassian.net/browse/AUT-11979


examples/client/Locomotion/src/context/newRideContext/index.tsx
```diff
@@ -884,11 +884,11 @@ const RidePageContextProvider = ({ children }: {
   });
 
 
-  const searchAddress = useCallback(async (searchText: string) => {
```
OmerGery, line 887: no need useCallback?

OmerGery (approved):

PR #695, opened 2023-02-17
Add distance filter on map selection

dorshay6:
## ℹ️ Change description:
In some cases, the map updates slightly between screens and makes the geocoding run again, overwriting the user address. This should solve it

<p>
  
</p>
</details>


avivbarniv (commented):
it might be a bit tough to calculate the distance every time

PR #744, opened 2023-05-11
add reCaptcha

OmerGery:


examples/client/Locomotion/src/pages/AuthScreens/StartScreen/index.js
```diff
@@ -21,14 +25,36 @@ import { INITIAL_USER_STATE } from '../AuthLoadingScreen';
...
     setUser(INITIAL_USER_STATE);
-    navigationService.navigate(MAIN_ROUTES.PHONE);
+    if (Config.CAPTCHA_KEY) {
+      if (recaptchaRef.current) {
+        recaptchaRef.current.open();
+      }
+    } else {
+      navigationService.navigate(MAIN_ROUTES.PHONE);
+    }
   };
```
oriefrati: use 'useCallback' with Config.CAPTCHA_KEY dependency

examples/client/Locomotion/src/services/network.js
```diff
@@ -68,11 +68,14 @@ class Network {
...
-        const accessToken = await Auth.getAT(this.axios);
+        const [accessToken, captchaToken] = await Promise.all([
+          Auth.getAT(this.axios),
+          Config.CAPTCHA_KEY && Auth.getCaptchaToken(),
+        ]);
         this.axios.defaults.headers.common.Authorization = accessToken ? `Bearer ${accessToken}` : accessToken;
-
         this.axios.defaults.headers.common['x-loco-ds-id'] = operationId;
         this.axios.defaults.headers.common['x-loco-op-id'] = operationId;
+        this.axios.defaults.headers.common['x-captcha-token'] = captchaToken;
```
oriefrati: do we want to add it for every request?
as I understand its generate it only on the login right?

OmerGery: adjusted it so only if no access token we add it , that way it will be used only on the relevant endpoints

examples/client/Locomotion/src/pages/AuthScreens/StartScreen/index.js
```diff
@@ -21,16 +26,38 @@ import { INITIAL_USER_STATE } from '../AuthLoadingScreen';
...
 
 const StartScreen = () => {
   const { setUser } = useContext(UserContext);
   const { getSettingByKey } = Settings.useContainer();
   const [webViewWindow, setWebViewWindow] = useState(null);
-  const nextScreen = () => {
-    setUser(INITIAL_USER_STATE);
-    navigationService.navigate(MAIN_ROUTES.PHONE);
+  const recaptchaRef = useRef(null);
+  const [captchaToken, setCaptchaToken] = useState('');
```
oriefrati: set null as default

examples/client/Locomotion/src/services/network.js
```diff
@@ -68,9 +68,14 @@ class Network {
...
         ]);
         this.axios.defaults.baseURL = baseURL;
-        const accessToken = await Auth.getAT(this.axios);
+        const [accessToken, captchaToken] = await Promise.all([
+          Auth.getAT(this.axios),
+          Config.CAPTCHA_KEY && Auth.getCaptchaToken(),
+        ]);
+        if (!accessToken && captchaToken) {
+          this.axios.defaults.headers.common['x-captcha-token'] = captchaToken;
+        }
```
oriefrati: you need to check only if captchaToken exists (it doesnt matter if there is accessToken or not)
and I think that once you insert token to the local storage, it stays forever
im don't see where you delete it

oriefrati (approved):

PR #760, opened 2023-06-26
Draglist - align with main lib again after issue resolved

OmerGery:
https://github.com/fivecar/react-native-draglist/issues/14 the lib owner fixed the issue so we can return from the fork


ariel-weiss (approved):
Nice 😄

PR #764, opened 2023-07-10
Add a fallback to use IP to get county code

roei-autofleet:
## 📝 Jira tickets:
https://autofleet.atlassian.net/browse/AUT-15397

## ℹ️ Change description:
add a fallback to use IP to get county code


examples/client/Locomotion/src/services/MccMnc.js
```diff
@@ -12,6 +12,16 @@ const getMobileIsoCode = async () => {
   const isoCode = await CarrierInfo.isoCountryCode();
   return isoCode.toUpperCase();
 };
+export const getCountryCodeByIP = async () => {
+  try {
+    const response = await fetch('https://ipapi.co/json/');
```
sephina: extract to const (maybe env var?)

examples/client/Locomotion/src/services/MccMnc.js
```diff
@@ -12,6 +12,16 @@ const getMobileIsoCode = async () => {
...
+    const data = await response.json();
+    return data?.country;
+  } catch (error) {
+    console.error('Unable to get country code by IP', error);
```
sephina: prefer mixpanel event

sephina (commented):
what is that api? can we count on its reliability?
also merge to version branch and not directly to master

roei-autofleet:
> what is that api? can we count on its reliability? also merge to version branch and not directly to master

Changed to the version
It's a pretty known free API that provide the location of the IP, this way we don't need to ask for location permissions to set the country code

sephina (approved):

PR #773, opened 2023-09-03
If price is undefined - return 0

ormiz:


lieblichaaron (approved):

ariel-weiss (changes requested):
please change target branch to the version branch

examples/client/Locomotion/src/context/newRideContext/utils.ts
```diff
@@ -165,11 +165,12 @@ export const formatStopPointsForEstimations = (requestStopPoints: any[]) => requ
   lng: sp.lng,
 }));
 
-export const getFormattedPrice = (priceCurrency: string | undefined, priceAmount: number) => {
+export const getFormattedPrice = (priceCurrency: string | undefined,
+  priceAmount: number | undefined) => {
```
ariel-weiss, line 169: instead of explicitly add undefined in the type, you can mark it as optional (with ? ) depends on why the parameter is undefined

avivbarniv (approved):

examples/client/Locomotion/src/context/newRideContext/utils.ts
ormiz: There are many options here, I could also set default value as 0, I just wanted to match it with priceCurrency

PR #789, opened 2023-10-16
change outstanding balance ui

avivbarniv:


examples/client/Locomotion/ios/Podfile
```diff
@@ -53,5 +53,10 @@ target 'Locomotion' do
         end
       end
     end
+    installer.pods_project.targets.each do |target|
+      target.build_configurations.each do |config|
+        config.build_settings['EXCLUDED_ARCHS[sdk=iphonesimulator*]'] = "arm64"
+      end
+    end
   end
```
ormiz: I have this local change too, would it work on intel-based Macs too?

ormiz (approved):

examples/client/Locomotion/ios/Podfile
avivbarniv: I will remove it

PR #791, opened 2023-11-09
hide price

hilat-autofleet:
https://autofleet.atlassian.net/browse/AUT-16921
hide the ride price when setting is demanding it.


avivbarniv (changes requested):
I think it's better to handle it as a setting in the settings context, there is a cache and it will work better

OmerGery (approved):

avivbarniv (approved):

PR #807, opened 2024-02-04
Aut 18615 switch demand source

ormiz:


examples/client/Locomotion/src/context/user/index.tsx
```diff
@@ -175,11 +175,17 @@ const UserContextProvider = ({ children }: { children: any }) => {
...
   const onLogin = async (phoneNumber: string, channel = 'sms') => {
     const demandSourceId = await AppSettings.getOperationId();
-    await loginApi({
+    const switchDemandSource = Config.SWITCH_DEMAND_SOURCE === 'true';
+    const response = await loginApi({
       phoneNumber,
       channel,
       demandSourceId,
+      switchDemandSource,
     });
```
OmerGery, line 197: const  { demandSourceId: responseDemandSourceId } = response;

OmerGery (approved):

examples/client/Locomotion/src/context/user/index.tsx
ormiz: Now it's selectedDemandSourceId, so no need to rename

examples/client/Locomotion/src/context/user/index.tsx
```diff
@@ -173,13 +173,27 @@ const UserContextProvider = ({ children }: { children: any }) => {
     return vertResponse.status === 'OK';
   };
 
+  const getAllowedDemandSourceIds = () => {
+    if (!Config.ALLOWED_DEMAND_SOURCE_IDS) {
+      return [];
+    }
+    const allowedDemandSourceIds = JSON.parse(Config.ALLOWED_DEMAND_SOURCE_IDS);
```
OmerGery: add try catch to the JSON parse + mixpanel event. 
As it ui only logic , it will help us to debug if the Config.ALLOWED is invalid json or undefined

PR #822, opened 2024-04-10
fix show price in future ride + don't display offline text in business account payment

OmerGery:


examples/client/Locomotion/src/Components/RideCard/index.tsx
OmerGery, line 1: future ride fix

roei-autofleet (approved):

PR #833, opened 2024-07-18
captcha hotfix

OmerGery:


examples/client/Locomotion/src/pages/Profile/Phone.js
```diff
@@ -49,10 +47,9 @@ const Phone = ({ navigation }) => {
   }, []);
 
 
-  const onVerifyCaptcha = async (token) => {
-    Mixpanel.setEvent('Captcha Verified successfully', { token });
```
OmerGery, line 53: token is a reserved word in mixpanel, caused this log to never be sent
https://docs.mixpanel.com/docs/data-structure/events-and-properties

examples/client/Locomotion/src/pages/Profile/Phone.js
```diff
@@ -80,6 +77,7 @@ const Phone = ({ navigation }) => {
       if (!isDevSettingOn()) {
         await AppSettings.destroy();
       }
+      await Auth.updateCaptchaToken(captchaToken);
```
OmerGery, line 80: https://github.com/Autofleet/locomotion/pull/833/files#diff-c05e09a619ce22ac8b2863115ae42f257ce102e5c7aa326a1f36740dd7f1b95fR78 is deleting the token from the storage. moved it to be after the destroy

atlowChemi (approved):

barspielberg (approved):

examples/client/Locomotion/src/pages/Profile/Phone.js
sephina: what are you doing awake on 7 am?

sephina (approved):

examples/client/Locomotion/src/pages/Profile/Phone.js
ariel-weiss: oww edgy

that's why we need typescript everywhere

OmerGery: Dreaming of captcha issues

PR #873, opened 2025-04-17
[AUT-24091] Fix service sorting logic, sort by `serviceAvailabilitiesNumber` -> sort by `priority`

Davidi-74:


examples/client/Locomotion/src/context/newRideContext/index.tsx
```diff
@@ -373,17 +373,13 @@ const RidePageContextProvider = ({ children }: {
     });
     const formattedServices = services.map((service) => {
       const estimationForService = estimationsMap[service.id];
-      return formatEstimationsResult(service, estimationForService, tags);
+      return formatEstimationsResult(service, estimationForService, tags, !!ride.scheduledTo);
     });
+
     return formattedServices.sort((a, b) => {
```
danielmar121, line 379: Can do
```suggestion
    return formattedServices.sort((a, b) => b.unavailable? -1 : a.priority - b.priority,
```

danielmar121, line 379: Can you add a unit test for this sorting function?

examples/client/Locomotion/src/context/newRideContext/utils.ts
```diff
@@ -130,15 +130,25 @@ export const latLngToAddress = async (lat: string, lng: string) => {
...
+export const formatEstimationsResult = (
+  service: any,
+  estimationResult: any,
+  tags: any,
+  isFutureRide: boolean,
+) => {
   const estimation = estimationResult || {};
+
+  const eta = estimation.minPickupEta;
+  const isHighEtaAsapRide = !!estimation.highEtaAsapRide;
```
danielmar121, line 142: What if this is an empty object? I am not familiar with this field

danielmar121 (approved):
LGTM

examples/client/Locomotion/src/context/newRideContext/index.tsx
Davidi-74: Not a big fan of ternary functions 😅
Mainly because they tend to bloat:
```
// starting out with something like
return formattedServices.sort((a, b) => b.unavailable ? -1 : a.priority - b.priority,);

// can easily turn to something like
return formattedServices.sort((a, b) => b.unavailable ? -1 : b.anotherCondition ? a.priority - b.priority : 1);
```

examples/client/Locomotion/src/context/newRideContext/utils.ts
Davidi-74: Generally it would crash long before it gets here, since `estimation` is being read multiple times. 
There is an error handler in `examples/client/Locomotion/src/context/newRideContext/index.tsx` that would catch it, but it's not great I agree

examples/client/Locomotion/src/context/newRideContext/index.tsx
Davidi-74: Talked with @lieblichaaron and @ariel-weiss offline, sadly we don't have any unit tests here

lieblichaaron (approved):

PR #877, opened 2025-05-13
Upgrade `OneSignal` from `v4 -> v5`, `hermes` black magic, refactor `one-signal.js -> one-signal.ts`

Davidi-74:
The reason for this PR is due to Apple's restriction of building apps w/ Xcode <16.x. 
Once building with Xcode 16.2 we cannot have dependencies that use Bitcode, like `OneSignal` v4 and the usage of `hermes` in our Podfile with the current `react-native` version.

Since we need to make a lot of changes to the way we implement `OneSignal`, I took the chance to refactor the service and convert it into TS.

### DIFF
--- a/examples/client/Locomotion/src/context/user/index.tsx
+++ b/examples/client/Locomotion/src/context/user/index.tsx
@@ -44,7 +44,6 @@ interface UserContextInterface {
   getUserFromServer: () => Promise<void>,
   locationGranted: boolean | undefined,
   setLocationGranted: Dispatch<SetStateAction<any>>,
-  updatePushToken: () => Promise<boolean | null>,
   deleteUser: () => Promise<boolean>,
   updateUser: (values: any) => Promise<any>,
   coupon: any | null,
@@ -68,7 +67,6 @@ export const UserContext = createContext<UserContextInterface>({
   getUserFromServer: async () => undefined,
   locationGranted: false,
   setLocationGranted: () => undefined,
-  updatePushToken: async () => false,
   deleteUser: async () => true,
   updateUser: async (values: any) => undefined,
   coupon: null,
@@ -118,33 +116,6 @@ const UserContextProvider = ({ children }: { children: any }) => {
     getUserFromStorage();
   }, []);
 
-  const updatePushToken = async () => {
-    const deviceState = await OneSignal.getDeviceState();
-    if (!deviceState) {
-      await updateUserInfo({ pushTokenId: null });
-      return null;
-    }
-
-    const { isSubscribed, pushToken, userId } = deviceState;
-
-    if (
-      user?.pushTokenId !== pushToken
-      || user?.pushUserId !== userId
-      || user?.isPushEnabled !== isSubscribed
-    ) {
-      await updateUserInfo({
-        pushTokenId: pushToken,
-        isPushEnabled: isSubscribed,
-        pushUserId: userId,
-      });
-    }
-    return true;
-  };
-
-  const verifyEmail = async () => {
-    await sendEmailVerification();
-  };
-
   const updateUserInfo = async (values: any, { updateServer = true } = {}) => {
     updateState(values);
     const newUser = updateServer ? await updateUserApi(values) : values;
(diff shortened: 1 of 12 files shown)


examples/client/Locomotion/src/context/user/index.tsx
```diff
@@ -12,7 +13,7 @@ import {
 import auth from '../../services/auth';
 import Mixpanel from '../../services/Mixpanel';
 import PaymentsContext from '../payments';
-import OneSignal from '../../services/one-signal';
+import { PushSettings } from '../../services/one-signal/types';
```
Elransh:
```suggestion
import type { PushSettings } from '../../services/one-signal/types';
```

examples/client/Locomotion/src/services/one-signal/index.ts
```diff
@@ -0,0 +1,176 @@
...
+    } catch (error) {
+      console.error('Error getting device state', error);
+      return null;
+    }
+  };
+
+  refreshPushSettings = async (): Promise<PushSettings | null> => {
+    try {
+      const pushSettings = await this.getPushSettings();
+      if (!pushSettings) return null;
```
danielmar121, line 78: We do not want to return an empty object or undefined? why null?

examples/client/Locomotion/src/services/one-signal/index.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (additionalData?.type && this.foregroundNotificationsHandlers[additionalData.type]) {
+      this.foregroundNotificationsHandlers[additionalData.type](additionalData);
+    }
+
+    notificationReceivedEvent.preventDefault();
+    notificationReceivedEvent.getNotification().display();
+  };
+
+  init = async (userId: string): Promise<PushSettings | null> => {
+    if (!Config.ONESIGNAL_APP_ID) return null;
```
danielmar121, line 140: why null?

examples/client/Locomotion/src/services/one-signal/index.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    if (pushSubscriptionId && pushToken) {
+      await this.updateServer(pushToken, pushSubscriptionId, isPushEnabled);
+    }
+  };
+
+
+  registerOnServer = async (pushUserData: PushUserData): Promise<void> => {
+    const response = await updateUser(pushUserData);
+    console.log(response.data);
```
danielmar121: 😱

examples/client/Locomotion/src/services/one-signal/index.ts
```diff
@@ -0,0 +1,176 @@
...
+    OneSignal.initialize(Config.ONESIGNAL_APP_ID);
+    OneSignal.login(userId);
+
+    OneSignal.User.pushSubscription.optIn();
+
+    OneSignal.Notifications.addEventListener('click', this.handleNotificationClick);
+    OneSignal.Notifications.addEventListener('foregroundWillDisplay', this.handleForegroundNotificationClick);
+    OneSignal.User.pushSubscription.addEventListener('change', this.subscriptionObserverHandler);
+
+    if (Platform.OS === 'ios') {
```
danielmar121: Let put the ios string in const in the start of the file

examples/client/Locomotion/src/services/one-signal/index.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (additionalData?.type && this.notificationsHandlers[additionalData.type]) {
+      this.notificationsHandlers[additionalData.type](additionalData);
+    }
+  };
+
+  handleForegroundNotificationClick = (
+    notificationReceivedEvent: NotificationWillDisplayEvent,
+  ): void => {
+    const { notification } = notificationReceivedEvent;
+    const additionalData = notification.additionalData as NotificationAdditionalData;
```
danielmar121, line 112: NotificationWillDisplayEvent does not contain additionalData as NotificationAdditionalData? We must to do that?

examples/client/Locomotion/src/services/one-signal/index.ts
Davidi-74: awkward...

examples/client/Locomotion/src/services/one-signal/index.ts
Davidi-74: It does contain `additionalData`, but `additionalData` is defined as an object in which anything can be set. Instead of defining it as something like this:
```
type AdditionalData {
  [key: string]: any;
}
```

The just defined this as an `object`. That's why I created `NotificationAdditionalData` with `type`, which is what we might pass in `additionalData`.

examples/client/Locomotion/src/services/one-signal/index.ts
Davidi-74: discussed offline

examples/client/Locomotion/src/services/one-signal/index.ts
Davidi-74: discussed offline

danielmar121 (approved):

examples/client/Locomotion/src/services/one-signal/index.ts
```diff
@@ -0,0 +1,171 @@
...
+    OneSignal.login(userId);
+
+    OneSignal.User.pushSubscription.optIn();
+
+    OneSignal.Notifications.addEventListener('click', this.handleNotificationClick);
+    OneSignal.Notifications.addEventListener('foregroundWillDisplay', this.handleForegroundNotificationClick);
+    OneSignal.User.pushSubscription.addEventListener('change', this.subscriptionObserverHandler);
+
+    if (Platform.OS === IOS) {
+      const permission = await OneSignal.Notifications.requestPermission(true);
```
ofekrotem: passing true will refer the user to open settings if they declined push notification permissions, see how we fixed it for the driver app:
https://github.com/Autofleet/driver-app/pull/1898

danielmar121: Updated

danielmar121 (approved):

PR #878, opened 2025-05-30
Fixed MewRidePageContext typo

stevenwithaph:
## ℹ️ Change description:

Noticed this typo as I going through the code, don't think it was supposed to be called "MewRidePageContext"


Davidi-74 (approved):
Hey Steven!
Thank you for taking the time to create the PR, I'll make sure to include it in the next version :fire:

PR #885, opened 2025-07-31
[RN 0.74] Replace `react-native-smooth-pincode-input` with `react-native-confirmation-code-field`

Davidi-74:
# Summary
This PR deprecates `react-native-smooth-pincode-input`, which is no longer supported in RN74. Instead, `react-native-confirmation-code-field` will be added.

It also contains a fix in `examples/client/Locomotion/src/Components/Thumbnail/index.js`, that uses `propTypes`, which is deprecated in RN74.

In addition, there are 2 QOL changes:
- Removal of `import/prefer-default-export` - as we're moving to a more modern way of using named exports instead of default exports.
- Added Error color to theme.


ormiz (approved):

ariel-weiss (approved):
lgtm

PR #889, opened 2025-08-03
Fix Android excessive CPU usage due to animated markers

Davidi-74:
# Summary
Custom animated markers cause excessive CPU usage on Android devices. This can be solved by setting `tracksViewChanges` to `false`, which disables the Marker automatically animating its position, which is CPU intensive.
Instead, we manually animate the Markers position using `AnimationRegion`.
In addition, there are some TS improvements and typo fixes.

### DIFF
--- a/examples/client/Locomotion/src/Components/AvailabilityVehicle/index.tsx
+++ b/examples/client/Locomotion/src/Components/AvailabilityVehicle/index.tsx
@@ -2,9 +2,9 @@ import React, {
   useContext, useEffect, useState, useRef, useCallback,
 } from 'react';
 import {
-  Platform,
-} from 'react-native';
-import { MarkerAnimated, AnimatedRegion } from 'react-native-maps';
+  MarkerAnimated, AnimatedRegion, MapMarker, LatLng,
+} from 'react-native-maps';
+import { StyleProp } from 'react-native';
 import { Context as ThemeContext } from '../../context/theme';
 import SvgIcon from '../SvgIcon';
 import carIcon from '../../assets/map/Autofleet_Car_Icon.svg';
@@ -20,17 +20,28 @@ interface AvailabilityVehicleProps {
   id: string;
 }
 
+interface SvgStyle {
+  color: string,
+  transform?: { rotate: string }[]
+}
+
+/** toValue is ignored in favor of lat/lng, but required by TimingAnimationConfig interface */
+const ANIMATED_VALUE_PLACEHOLDER = 1;
+
+/** Zero delta maintains current map zoom level during marker animation */
+const MAINTAIN_CURRENT_ZOOM = 0;
+
 const DURATION = 5000;
 
 const areEqual = (
   prev: AvailabilityVehicleProps,
   next: AvailabilityVehicleProps,
 ) => prev.id === next.id
-  && prev.location.lat === next.location.lat
-  && prev.location.lng === next.location.lng
+&& prev.location.lat === next.location.lat
+&& prev.location.lng === next.location.lng
   && prev.location.bearing === next.location.bearing;
 
-const insureNumberType = (v: string | number) => {
+const ensureNumberType = (v: string | number) => {
   if (typeof v === 'string') {
     return parseFloat(v);
   }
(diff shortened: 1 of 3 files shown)


examples/client/Locomotion/src/Components/AvailabilityVehicle/index.tsx
```diff
@@ -20,17 +19,28 @@ interface AvailabilityVehicleProps {
   id: string;
 }
 
+interface SvgStyle {
+  color: string,
+  transform?: { rotate: string }[]
+}
+
+/** toValue is ignored in favor of lat/lng, but required by TimingAnimationConfig interface */
+const ANIMATED_VALUE_PLACEHOLDER = 1;
```
ariel-weiss: if the description is related to that const, I don't understand how

edit: OK i kept reading and now i understand 😉 if you can make it clearer it'll be great, not blocker

examples/client/Locomotion/src/Components/AvailabilityVehicle/index.tsx
```diff
@@ -42,34 +52,34 @@ const AvailabilityVehicle = ({
...
-    latitudeDelta: 0.1,
-    longitude: insureNumberType(location.lng),
-    longitudeDelta: 0.1,
-  }));
+  const markerRef = useRef<MapMarker>(null);
+  const locationAnimationRef = useRef<AnimatedRegion>(
+    new AnimatedRegion({
+      latitude: ensureNumberType(location.lat),
+      longitude: ensureNumberType(location.lng),
+      latitudeDelta: MAINTAIN_CURRENT_ZOOM,
```
ariel-weiss: changed 0.1 to 0 on purpose?

examples/client/Locomotion/src/Components/AvailabilityVehicle/index.tsx
```diff
@@ -42,34 +52,34 @@ const AvailabilityVehicle = ({
...
-        }).start();
-      }
+      locationAnimationRef.current.timing({
+        toValue: ANIMATED_VALUE_PLACEHOLDER,
+        duration: DURATION,
+        longitudeDelta: MAINTAIN_CURRENT_ZOOM,
+        latitudeDelta: MAINTAIN_CURRENT_ZOOM,
+        useNativeDriver: false,
+        latitude: location.lat,
+        longitude: location.lng,
```
ariel-weiss: no need to `ensureNumberType` here?

examples/client/Locomotion/src/Components/AvailabilityVehicle/index.tsx
```diff
@@ -42,34 +52,34 @@ const AvailabilityVehicle = ({
...
+      }).start();
     }
   }, [location]);
```
ariel-weiss: can be
```suggestion
  }, [location.lat, location.lng]);
```
or we need other location props?

examples/client/Locomotion/src/Components/AvailabilityVehicle/index.tsx
Davidi-74: good catch, not intentional

examples/client/Locomotion/src/Components/AvailabilityVehicle/index.tsx
Davidi-74: updated to `TO_VALUE_PLACEHOLDER`

examples/client/Locomotion/src/Components/AvailabilityVehicle/index.tsx
Davidi-74: was this way before, but I think adding it makes sense

examples/client/Locomotion/src/Components/AvailabilityVehicle/index.tsx
Davidi-74: Technically, "it doesn't matter" (it does matter because using an object as a useEffect dep is generally wrong), because the `location` object reference changes every time we fetch the vehicles from the API, so this works.
Changed to what you suggested because it makes more sense - but the correct solution here would be to pass `...location`, aka `lat`, `long` and `bearing` instead of the entire object. Sadly, I don't have time for that ATM

ariel-weiss: oki doki

examples/client/Locomotion/src/Components/AvailabilityVehicle/index.tsx
```diff
@@ -20,17 +20,28 @@ interface AvailabilityVehicleProps {
...
+interface SvgStyle {
+  color: string,
+  transform?: { rotate: string }[]
+}
+
+/** toValue is ignored in favor of lat/lng, but required by TimingAnimationConfig interface */
+const TO_VALUE_PLACEHOLDER = 1;
+
+/** Zero delta maintains current map zoom level during marker animation */
+const MAINTAIN_CURRENT_ZOOM = 0;
```
ariel-weiss, line 32: no need here 0.1?

ariel-weiss (approved):

examples/client/Locomotion/src/Components/AvailabilityVehicle/index.tsx
Davidi-74: nope, we should keep it at 0 for the animation

PR #893, opened 2025-08-14
AUT-25412: Add `External` payment method

galinareznikk:
The currently supported payment methods are cash, offline, and credit card.
The task is to add an external payment method.

This includes updating all payment screens where payment options are displayed, such as the main flow when ordering a ride, scheduling future rides, etc.

The external payment method should not be displayed in the payment section of the hamburger menu.

This option is already supported in the Control Center through a configuration flag.
In Locomotion, it will be presented according to the service definition.

### DIFF
--- a/examples/client/Locomotion/src/Components/RideCard/index.tsx
+++ b/examples/client/Locomotion/src/Components/RideCard/index.tsx
@@ -13,6 +13,7 @@ import StopPointsVerticalView from '../StopPointsVerticalView';
 import { getFormattedPrice, isPriceEstimated, convertTimezoneByLocation } from '../../context/newRideContext/utils';
 import cashIcon from '../../assets/cash.svg';
 import offlineIcon from '../../assets/offline.svg';
+import placeholderIcon from '../../assets/placeholder-payment.svg';
 import { PAYMENT_METHODS } from '../../pages/Payments/consts';
 import PaymentContext from '../../context/payments';
 import SettingContext from '../../context/settings';
@@ -31,6 +32,7 @@ interface CardComponentProps {
 const CardComponent = ({ paymentMethod, businessAccountId }: CardComponentProps) => {
   const isCash = PAYMENT_METHODS.CASH === paymentMethod.id;
   const isOffline = PAYMENT_METHODS.OFFLINE === paymentMethod.id;
+  const isExternal = PAYMENT_METHODS.EXTERNAL === paymentMethod.id;
   const {
     offlinePaymentText,
     loadOfflinePaymentText,
@@ -51,6 +53,10 @@ const CardComponent = ({ paymentMethod, businessAccountId }: CardComponentProps)
     } if (isOffline) {
       return offlinePaymentText;
     }
+    if (isExternal) {
+      return i18n.t('payments.external');
+    }
+
     return paymentMethod.name;
   };
 
@@ -61,7 +67,11 @@ const CardComponent = ({ paymentMethod, businessAccountId }: CardComponentProps)
     if (isOffline) {
       return offlineIcon;
     }
+    if (isExternal) {
+      return placeholderIcon;
+    }
   };
+
   return (
     <TextRowWithIcon
       text={getText() || ''}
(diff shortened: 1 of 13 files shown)


itayankri (approved):

examples/client/Locomotion/src/Components/RideCard/index.tsx
```diff
@@ -61,7 +67,11 @@ const CardComponent = ({ paymentMethod, businessAccountId }: CardComponentProps)
     if (isOffline) {
       return offlineIcon;
     }
+    if (isExternal) {
+      return placeholderIcon;
```
Davidi-74: I think, instead of doing this, we can:
```suggestion
      return null;
```

And then, in `TextRowWithIcon`, do something like:
```
if (!icon) return <View />;
```

Or:
```
if (!icon) return null;
```

examples/client/Locomotion/src/pages/Payments/externalPaymentMethod.ts
```diff
@@ -0,0 +1,22 @@
...
+  isExpired: false,
+  lastFour: '',
+  name: i18n.t('payments.external'),
+  stripeId: '',
+  updatedAt: new Date(),
+  deletedAt: null,
+  outstandingBalance: { amount: 0, currency: 'USD' },
+};
+
+export default externalPaymentMethod;
```
Davidi-74: Small, but lets not use a default export:
```suggestion
export const externalPaymentMethod : PaymentMethodInterface = {
  brand: 'generic',
  createdAt: new Date(),
  customerId: '',
  expiresAt: new Date(2100, 9, 9),
  hasOutstandingBalance: false,
  id: PAYMENT_METHODS.EXTERNAL,
  isDefault: false,
  isExpired: false,
  lastFour: '',
  name: i18n.t('payments.external'),
  stripeId: '',
  updatedAt: new Date(),
  deletedAt: null,
  outstandingBalance: { amount: 0, currency: 'USD' },
};
```

examples/client/Locomotion/src/Components/RideCard/index.tsx
ariel-weiss: @Davidi-74 but then you won't have the placeholder icon , why like this/?

ariel-weiss: @galinareznikk I just noticed the placeholder icon is actually empty, is this what Product wanted?

examples/client/Locomotion/src/pages/ActiveRide/RideDrawer/RideOptions/RideButtons/PaymentButton/index.tsx
```diff
@@ -95,7 +100,7 @@ const PaymentButton = ({
     : i18n.t('bottomSheetContent.ride.promoText'));
 
   const loadPromoButton = () => {
-    if (isCashPaymentMethod({ id }) || isOfflinePaymentMethod({ id })) {
+    if (isCashPaymentMethod({ id }) || isOfflinePaymentMethod({ id }) || isExternalPaymentMethod({ id })) {
```
ariel-weiss, line 103: `isCashPaymentMethod` already has `isExternalPaymentMethod` in it, it seems

ariel-weiss (approved):
nice 🔥 💰

examples/client/Locomotion/src/pages/Payments/externalPaymentMethod.ts
galinareznikk: fixed

examples/client/Locomotion/src/Components/RideCard/index.tsx
galinareznikk: I asked Ido about this, and he told me to use the default payment icon
`import { PaymentIcon } from 'react-native-payment-icons'`
So there’s no need for a placeholder icon.

examples/client/Locomotion/src/pages/Payments/externalPaymentMethod.ts
```diff
@@ -0,0 +1,20 @@
+import { PaymentMethodInterface } from 'context/payments/interface';
+import i18n from 'i18next';
+import { PAYMENT_METHODS } from './consts';
+
+export const externalPaymentMethod : PaymentMethodInterface = {
+  brand: 'generic',
+  createdAt: new Date(),
+  customerId: '',
+  expiresAt: new Date(2100, 9, 9),
```
Davidi-74, line 9: Why? 😅

Davidi-74 (approved):

Davidi-74 (approved):

PR #900, opened 2025-09-11
Fix `MainCardForm` causing a screen freeze

Davidi-74:
# Summary
This PR includes a fix to a crash that occurs when trying to switch a country in the `NewCreditForm` screen. 
The crash happens due to a crash with Stripe's `MainCardForm` and React Native's `KeyboardAvoidingView`.
Since that screen is fairly compact, `KeyboardAvoidingView` can be removed as the inputs shouldn't go under the Keyboard anyway.

|Before|After|
|---|---|
|Simulator Screenshot - iPhone 16 Pro - 2025-09-11 at 18 50 07|Simulator Screenshot - iPhone 16 Pro - 2025-09-11 at 18 48 44|

### DIFF
--- a/examples/client/Locomotion/src/Components/NewCreditForm/index.js
+++ b/examples/client/Locomotion/src/Components/NewCreditForm/index.js
@@ -72,7 +72,7 @@ const NewCreditForm = ({ onDone, canSkip = false, PageText }) => {
           }}
           style={{
             width: '100%',
-            height: 170,
+            height: 300,
             border: 0,
           }}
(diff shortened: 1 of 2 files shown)


dorshay6 (approved):

atlowChemi (commented):
LGTM, but what happens when phone is horizontal? (if that can even happen)

Davidi-74:
> LGTM, but what happens when phone is horizontal? (if that can even happen)

Horizontal mode isn't supported

atlowChemi (approved):

PR #910, opened 2025-11-23
fix payment method selector

asher-mentzer:


examples/client/Locomotion/src/Components/CardRow/index.tsx
```diff
@@ -135,22 +135,22 @@ const CardRow = (paymentMethod: any) => {
   }, [paymentMethod]);
 
   const prefix = paymentMethod.testIdPrefix || '';
-  const { paymentMethodId, addNew } = paymentMethod;
+  const { id, addNew } = paymentMethod;
   const isSpecialMethod = [
     PAYMENT_METHODS.OFFLINE,
     PAYMENT_METHODS.CASH,
     PAYMENT_METHODS.EXTERNAL,
-  ].includes(paymentMethodId);
```
galinareznikk, line 143: Why do we compare an ID with an array of payment methods? @Davidi-74

galinareznikk: @asher-mentzer if you verified that this is indeed ‘cash’, then I guess it’s fine - but the naming is very weird.

galinareznikk (approved):

PR #912, opened 2025-12-10
when future ride becomes active - treat as asap

sephina:
## 📝 Jira tickets:
https://autofleet.atlassian.net/browse/AF-4927

### DIFF
--- a/examples/client/Locomotion/src/context/futureRides/index.tsx
+++ b/examples/client/Locomotion/src/context/futureRides/index.tsx
@@ -4,17 +4,20 @@ import React, {
   useEffect,
   Dispatch,
   useContext,
+  useRef,
 } from 'react';
 import useInterval from '../../lib/useInterval';
 import { RideInterface } from '../newRideContext';
 import * as futureRideApi from './api';
 import { UserContext } from '../user';
+import { RIDE_STATES, RIDE_FINAL_STATES } from '../../lib/commonTypes';
 
 interface FutureRidesContextInterface {
     futureRides: RideInterface[];
     newFutureRide: RideInterface | null;
     setNewFutureRide: Dispatch<RideInterface | null>;
     loadFutureRides: () => Promise<void>;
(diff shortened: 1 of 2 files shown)


examples/client/Locomotion/src/context/newRideContext/index.tsx
```diff
@@ -624,6 +625,20 @@ const RidePageContextProvider = ({ children }: {
...
+        const formattedRide = await formatRide(transitionedRide);
+        setRide(formattedRide);
+
+        const screenFunction = RIDE_STATES_TO_SCREENS[formattedRide.state || ''];
+        if (screenFunction) {
+          screenFunction(formattedRide);
+        }
+      });
+    }
+  }, [onFutureRideTransition]);
```
tyemini, line 640: Shouldn't this effect run only when we initialize the new ride component?

examples/client/Locomotion/src/context/futureRides/index.tsx
```diff
@@ -4,34 +4,74 @@ import React, {
...
+        activatedRideIds.current.delete(id);
+      }
+    });
+    const activeRides = rides.filter(
+      (ride: RideInterface) => !RIDE_FINAL_STATES.includes(ride.state || ''),
+    );
+    setFutureRides(activeRides);
+  };
+
+  const setOnFutureRideTransition = (callback: (ride: RideInterface) => void) => {
```
tyemini, line 73: If multiple components call this, they'll overwrite each other, no?

sephina: the assumption is that a user can have only 1 active ride at a time. because if he has few how do we know which on e is the real active one?

examples/client/Locomotion/src/context/newRideContext/index.tsx
sephina: setOnFutureRideTransition doesn't really change

tyemini (approved):

PR #919, opened 2026-03-26
Upgrade to 0.76

iftah-saar:
Jira ticket: https://autofleet.atlassian.net/browse/AF-6886

# React Native Upgrade: 0.74.7 → 0.76.9

**Branch:** \`upgrade-to-0.76\`
**Scope:** 56 files changed, 22 commits (upgrade-only — lint/refactoring moved to separate branch \`lint-and-typescript-migration\`)

---

### 1. React Native 0.74.7 → 0.76.9
- Core framework upgrade with React 18.2.0 → 18.3.1
- All \`@react-native/*\` packages bumped to 0.76.9

### 2. New Architecture (Fabric) Enabled
- **Android:** \`newArchEnabled=true\` in \`gradle.properties\` — required for \`react-native-gesture-handler\` 2.30.x
- **iOS:** Removed legacy \`sourceURLForBridge:\` and \`concurrentRootEnabled\` from \`AppDelegate.mm\`

### 3. Android Build System Overhaul
- **Kotlin:** 1.9.22 → 2.0.21
- **Gradle wrapper:** 8.9 → 8.11.1, **AGP:** 8.7.0 → 8.6.0, **NDK:** 26.1 → 27.1
- Migrated \`app/build.gradle\` from legacy \`apply plugin:\` to \`plugins {}\` DSL
- Removed deprecated build config: APK splits, \`jscFlavor\`, Hermes/JSC conditional logic, Flipper
- **settings.gradle** restructured with \`pluginManagement\`; removed manual \`include ':library'\` declarations
- **Root build.gradle fix:** Added \`classpath("com.facebook.react:react-native-gradle-plugin")\` for library module compatibility
- **Root settings.gradle fix:** Added root-level \`includeBuild\` for dependency substitution

### 4. iOS Build Simplification
- Removed ~140 lines of Hermes bitcode stripping logic from \`Podfile\`
- Removed explicit \`ENABLE_BITCODE = 'NO'\` and React-Codegen build settings overrides
- Platform target changed from hardcoded \`'13.4'\` to dynamic \`min_ios_version_supported\`
- Simplified \`post_install\` block to just \`react_native_post_install()\`

### 5. Ruby 3.2.2 → 4.0.2
- CocoaPods constraint updated: \`'>= 1.13', '< 1.15'\` → \`'>= 1.15', '< 1.17'\`

### 6. Stripe Upgrade 0.35.1 → 0.59.2
- **New patch** (\`@stripe+stripe-react-native+0.59.2.patch\`): Adds missing Fabric codegen stubs (Java view manager delegates/interfaces for Stripe's native components: AddToWalletButton, CardField, CardForm, etc.)
- **Why:** Stripe 0.59.2 predates stable New Architecture support and doesn't run codegen for its native modules. This is the same pattern as \`RCTThirdPartyFabricComponentsProvider\` for other third-party libs.

### 7. react-native-modal → CompatModal Wrapper
- Created new \`CompatModal\` component using React Native's native \`Modal\`
- Fixes Android popup rendering issues under RN 0.76
- All 15 popups migrated
- **GenericError** special case: Uses native \`Alert.alert()\` on Android

### 8. react-native-svg 12.3 → 15.11 + SVG Country Flags
- **New \`CountryFlag\` component**: Renders SVG flags via \`country-flag-icons\` instead of emoji
- **Why:** iOS TextKit2 in RN 0.76 breaks Regional Indicator emoji sequences for country flags
- **Metro config** updated with custom \`resolveRequest\` to intercept flag imports

### 9. react-native-gesture-handler 2.16.2 → 2.30.0
- Primary driver for enabling \`newArchEnabled=true\`

### 10. Skeleton Library Swap
- Removed \`react-native-skeleton-placeholder\` → Added \`react-native-reanimated-skeleton\`
- **Why:** \`react-native-skeleton-placeholder\` depends on deprecated APIs and crashes at runtime under New Architecture. \`react-native-reanimated-skeleton\` is a drop-in replacement (nearly identical API, only the import changes).
- \`react-native-reanimated\` updated: 3.15.0 → 3.16.7

### 11. Removed Deprecated Packages
- \`react-native-exit-app\`, \`@react-native-masked-view/masked-view\`
- Removed \`jetify\` from postinstall, removed Flipper

### 12. RN 0.76.9 Patch
Two Hermes crash fixes + Fabric component registration:
- **\`promiseRejectionTrackingOptions.js\`**: \`Error.stack\` getter may throw on Hermes with an invalid receiver — wrapped in try-catch
- **\`ExceptionsManager.js\`**: Same issue in \`reactConsoleErrorHandler\` — wrapped \`firstArg?.stack\` access in try-catch
- **\`RCTThirdPartyFabricComponentsProvider\`**: Registers Lottie, GestureHandler, SafeArea, Screens, WebView with the Fabric component registry (these libraries don't self-register)

### 13. iOS Privacy Manifest
- \`PrivacyInfo.xcprivacy\` declares \`NSPrivacyAccessedAPITypes\` for APIs our own code accesses: UserDefaults, FileTimestamp, SystemBootTime, DiskSpace (required since iOS 17)
- Data collection disclosures (DeviceID, UserID, tracking domains) are intentionally omitted — these are already declared by Mixpanel/AppsFlyer SDK manifests and Apple merges them automatically

### 14. Bug Fixes
- **\`geo.js\`**: Geolocation timeout increased from 2 s → 10 s to reduce false timeout errors
- **\`newRideContext\`**: Skip \`initSps()\` when on \`ADDRESS_SELECTOR\` bottom sheet page to prevent unwanted stop-point resets

---

## What's NOT in this PR

Lint, formatting, and TypeScript migration changes have been moved to a separate branch: **\`lint-and-typescript-migration\`** (to be merged after this PR). That branch includes:
- propTypes → TypeScript interfaces (50+ components)
- defaultProps → destructured default parameters
- Function declaration style conversions
- New \`.d.ts\` type definition files
- ESLint rule compliance fixes

---

## Build Verification
- **iOS:** BUILD SUCCEEDED
- **Android:** BUILD SUCCEEDED


Davidi-74:
@iftah-saar The upgrade guide is nice, but the PR itself is really noisy, and reading an upgrade guide is not the same as going through the code, as the code can change.

This is a very sensitive upgrade, and 90% of the changes are related to formatting. I understand that this is a side effect of upgrading ESLint configs, but I don't think it should be included in a React Native version upgrade. Even if it was - it should be split to another PR.

Can we disable `react/function-component-definition` and drop the formatting changes please? What would be the overhead of that?
In any case, I think switching from arrow functions to function declarations requires more internal discussion.

examples/client/Locomotion/ios/Locomotion/Info.plist
```diff
@@ -82,5 +82,7 @@
 	<string>Light</string>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
+	<key>RCTTextKit2Disabled</key>
+	<true/>
```
Davidi-74: Can't see this property in the [docs](https://developer.apple.com/documentation/bundleresources/information-property-list), does it exist?

examples/client/Locomotion/ios/Locomotion/PrivacyInfo.xcprivacy
Davidi-74, line 1: `NSPrivacyCollectedDataTypeTracking` is the only thing that exists in the [upgrade guide](https://react-native-community.github.io/upgrade-helper/?from=0.74.7&to=0.76.9), can you please specify why we need the other properties and their reasons?

examples/client/Locomotion/patches/@stripe+stripe-react-native+0.59.2.patch
Davidi-74, line 1: Can you please explain why we need this patch?

examples/client/Locomotion/patches/react-native+0.76.9.patch
Davidi-74, line 1: Same thing here

examples/client/Locomotion/src/Components/Marker/index.js
```diff
@@ -91,7 +90,6 @@ const StopPointMarker = ({
 
   return (
     <Marker
-      key={key}
```
Davidi-74, line 94: Why remove a component key?

examples/client/Locomotion/src/Components/PhoneNumberInput/CountryFlag.js
```diff
@@ -0,0 +1,51 @@
...
+    overflow: 'hidden',
+    borderRadius: 2,
+  },
+});
+
+const FlagImage = memo(({ countryCode, flagSize }) => {
+  const svg = flagSvgs[countryCode];
+  if (!svg) return null;
+  const height = 20;
+  const width = 30;
```
Davidi-74: Maybe hardcode these at the file level?

examples/client/Locomotion/src/popups/CancellationReasonsPopup/index.tsx
```diff
@@ -1,9 +1,9 @@
 import React, { useContext, useEffect, useState } from 'react';
-import Modal from 'react-native-modal';
 import { ScrollView } from 'react-native-gesture-handler';
+import Modal from '../../Components/CompatModal';
 import CloseButton from '../../Components/CloseButton';
-import CancellationReasonsProvider, { CancellationReasonsContext } from '../../context/cancellation-reasons';
-import RoundedButton from '../../Components/RoundedButton';
+import { CancellationReasonsContext } from '../../context/cancellation-reasons';
+import { CancellationReason } from '../../context/cancellation-reasons/interface';
```
Davidi-74: seems to be unused?

examples/client/Locomotion/metro.config.js
```diff
@@ -20,6 +21,27 @@ const config = {
   resolver: {
     assetExts: ['bin', 'txt', 'jpg', 'png', 'json', 'gif', 'webp', 'ttf', 'woff', 'woff2', 'otf'],
     sourceExts: ['js', 'jsx', 'ts', 'tsx', 'json', 'svg'],
+    /**
+     * Replace the emoji-based Flag component inside react-native-country-picker-modal
+     * (nested in react-native-phone-number-input) with a custom SVG implementation.
+     * This fixes country flag rendering on iOS with React Native 0.76 (TextKit2),
+     * where Regional Indicator Symbol emoji sequences no longer render correctly.
+     */
+    resolveRequest: (context, moduleName, platform) => {
```
Davidi-74: I think it would make more sense to move away from `react-native-phone-number-input` to something like `react-native-international-phone-number`, instead of doing a "surgical" hack

examples/client/Locomotion/package.json
```diff
@@ -82,15 +82,15 @@
     "react-native-payment-icons": "^1.0.11",
     "react-native-phone-number-input": "^2.1.0",
     "react-native-picker-select": "^8.0.4",
-    "react-native-reanimated": "~3.15.0",
+    "react-native-reanimated": "~3.16.7",
+    "react-native-reanimated-skeleton": "^1.6.0",
```
Davidi-74, line 86: Why is moving to a different Skeleton lib needed?

Davidi-74 (commented):
Let's also change the base branch here to the proper version, see https://github.com/Autofleet/locomotion/pull/911 and https://github.com/Autofleet/locomotion/pull/909

examples/client/Locomotion/ios/Locomotion/PrivacyInfo.xcprivacy
iftah-saar: The upgrade guide only shows the minimal template. The additional properties we declare are required by Apple for App Store submission based on the APIs our app actually uses:

- NSPrivacyCollectedDataTypes: We collect DeviceID and UserID (for Mixpanel/AppsFlyer analytics)
- NSPrivacyAccessedAPITypes: UserDefaults, FileTimestamp, SystemBootTime, DiskSpace — these are APIs that Apple requires disclosure for since iOS 17
- NSPrivacyTrackingDomains: Mixpanel and AppsFlyer domains

Without these, Apple will reject the app submission. These aren't from the RN upgrade guide — they're from our app's actual privacy requirements.

examples/client/Locomotion/ios/Locomotion/Info.plist
iftah-saar: You are right. It's not. I will try to find an alternative.

examples/client/Locomotion/patches/@stripe+stripe-react-native+0.59.2.patch
iftah-saar: The Stripe SDK (stripe-react-native 0.59.2) doesn't ship with pre-generated Fabric (New Architecture) component stubs. When building with newArchEnabled=true, the build expects generated Java view manager delegates and interfaces for Stripe's native components (AddToWalletButton, CardField, CardForm, etc.). The patch adds these generated codegen files so the Android build succeeds under the New Architecture. Without it, you get linker errors because the Fabric component registry can't find the Stripe view managers.

examples/client/Locomotion/patches/react-native+0.76.9.patch
iftah-saar: Two fixes:

- Hermes crash fix: promiseRejectionTrackingOptions.js — accessing Error.stack on Hermes can throw when the receiver is invalid. The patch wraps it in try-catch to prevent crashes on unhandled promise rejections.

- Third-party Fabric provider: Adds RCTThirdPartyFabricComponentsProvider files that register native components (Lottie, GestureHandler, SafeArea, Screens, WebView) with the Fabric component registry. Without this, these components fail to render under New Architecture.

examples/client/Locomotion/src/Components/Marker/index.js
iftah-saar: The key was being passed as a component prop to StopPointMarker and then forwarded to the <Marker> component. But key is a reserved React prop — it's used by React's reconciliation algorithm and is never passed down to the child component. Passing a key as a prop to a component does nothing (React strips it). The key should be placed on <StopPointMarker> at the call site (where it's rendered in a list), not forwarded internally. Removing it from the destructured props and the <Marker> is correct.

examples/client/Locomotion/metro.config.js
iftah-saar: That's a fair point in the long term, but swapping the phone input library is a feature change with its own scope - a different API, different styling, and different behavior that need testing. The Metro resolveRequest approach is isolated and only affects flag rendering, not the phone input UX. I'd suggest we keep the current approach for this upgrade PR and track the library swap as a separate task.

examples/client/Locomotion/package.json
iftah-saar: **react-native-skeleton-placeholder** is not compatible with React Native 0.76 / New Architecture (Fabric). It depends on deprecated APIs and crashes at runtime. **react-native-reanimated-skeleton** is a drop-in replacement that's actively maintained and works with Fabric. The API is nearly identical — only the import name changes.

examples/client/Locomotion/metro.config.js
Davidi-74: How is this different than the Skeleton lib though? The only difference is that one's blocked by a RN constraint and the other is an iOS constraint.

examples/client/Locomotion/patches/react-native+0.76.9.patch
Davidi-74: I find it hard to believe that patching RN makes more sense than tracking the RC of these that are probably caused by outdated dependencies. Can we understand what would be the tradeoff updating these?

examples/client/Locomotion/patches/@stripe+stripe-react-native+0.59.2.patch
Davidi-74: Same point as the RN patch; I find it hard to believe that an official package from Stripe doesn't support Fabric properly.

examples/client/Locomotion/ios/Locomotion/PrivacyInfo.xcprivacy
Davidi-74: IIRC these are already automatically added by AppsFlyer and Mixpanel, we do not need to specify these policies explicitly and it could lead to submission rejections.

examples/client/Locomotion/metro.config.js
iftah-saar: **Skeleton**: react-native-skeleton-placeholder is completely broken with RN 0.76 New Architecture - it won't build at all. There's no way to keep it; a library swap is the only option.

**Metro hack**: react-native-phone-number-input works fine with RN 0.76. The only problem is emoji flag rendering inside a deeply nested dependency (react-native-country-picker-modal). The library itself is fully functional.

Swapping to react-native-international-phone-number to fix a rendering issue in a nested component would mean:

Different API across all usage sites of the phone input
Different UI/UX and validation behavior that needs thorough testing
Broad, risky change for a narrow, isolated problem
The Metro resolveRequest replaces exactly one component (Flag) in exactly one nested package, with zero impact on phone input behavior. That's a much smaller blast radius than a full library swap.

The skeleton swap was forced. The Metro approach was a deliberate choice to keep the change minimal. They're not really analogous.

Davidi-74: The package is used in one component, and the UI/UX difference between them is marginal; Let's pass it through product.
Patching an abandoned package will just create more work in the long term, maybe even in the next version upgrade. We can do it in a different PR, but I don't think we should keep this long run.

examples/client/Locomotion/ios/Locomotion/PrivacyInfo.xcprivacy
iftah-saar: You're right — removed NSPrivacyCollectedDataTypes, NSPrivacyTracking, and NSPrivacyTrackingDomains. Only NSPrivacyAccessedAPITypes stays, which covers API access from our own app code (UserDefaults/FileTimestamp/SystemBootTime/DiskSpace — triggered by AsyncStorage, reanimated, etc.). Data collection and tracking declarations are owned by the Mixpanel and AppsFlyer SDK manifests.

examples/client/Locomotion/patches/react-native+0.76.9.patch
iftah-saar: These aren't caused by a specific outdated dependency — it's a known long-standing bug in RN's error handling code on Hermes: [facebook/react-native#43636](https://github.com/facebook/react-native/issues/43636). The root cause (documented in [facebook/hermes#1496](https://github.com/facebook/hermes/issues/1496)) is that many libraries create error subclasses using function(){} rather than class extends Error, which produces objects where Hermes's Error. stack getter throws on access. RN 0.77.3 and 0.78.3 ship the same unguarded code

examples/client/Locomotion/patches/@stripe+stripe-react-native+0.59.2.patch
iftah-saar: I cleaned up the patch — it was 11k lines because Gradle build-cache files were accidentally included. The actual change is 2 iOS stub files, 9 lines each (you can see them now in the patch).

Stripe does support Fabric properly for most components. The issue is specifically that ConnectAccountOnboardingView and PaymentMethodMessagingElementView are declared in their codegen spec (so the linker expects implementations), but Stripe hasn't shipped the .mm files for them — even in 0.63.0 (their latest). The stub pattern we use is Stripe's own convention; the comment in the file is copied from their own SDK.

Upgrading Stripe to 0.63.0 was the first thing I checked — same files are missing there.

examples/client/Locomotion/metro.config.js
iftah-saar: Fair enough — agreed they're the same category of problem. A few ways to handle it:

Option 1 — Remove the hack from this PR
Pull out CountryFlag.js, the metro.config.js resolveRequest, and country-flag-icons. Flag rendering on iOS remains broken in production until the library swap PR ships (the timeline depends on product scheduling).

Option 2 — Keep the hack as an interim measure, swap the library in a follow-up PR
The hack stays temporarily. We open a tracked issue now for the library swap and link it here so it doesn't get lost. No production regression in the meantime, clear exit path.

Option 3 — Do the full library swap in this PR
Replace react-native-phone-number-input with react-native-international-phone-number now, pass the UI diff through product in parallel, and merge once approved. Bigger scope for this PR,  but solves it properly in one shot.

Happy to go whichever direction you prefer. 

Here is how it looks on IOS with the flags bug:
Image

Davidi-74: Let's go with option 2 and open a ticket for the package update

examples/client/Locomotion/patches/react-native+0.76.9.patch
Davidi-74: I see, thanks for the explanation

examples/client/Locomotion/patches/@stripe+stripe-react-native+0.59.2.patch
Davidi-74: Cool, thanks for the explanation

package-lock.json
Davidi-74, line 1: unused

examples/client/Locomotion/src/services/geo.js
```diff
@@ -9,7 +9,7 @@ const ONE_SECOND = 1000;
 
 const DEFAULT_OPTIONS = {
   enableHighAccuracy: false,
-  timeout: 2 * ONE_SECOND,
+  timeout: 10 * ONE_SECOND,
```
Davidi-74, line 12: Why increase the timeout here?

examples/client/Locomotion/src/context/futureRides/index.tsx
Davidi-74, line 1: I think the changes in this file slipped from a past commit?

examples/client/Locomotion/src/context/newRideContext/index.tsx
Davidi-74, line 1: Same thing here

examples/client/Locomotion/src/context/futureRides/index.tsx
iftah-saar: It's because I pulled from master, and we're trying to merge this branch into [version_4.0.10](https://github.com/Autofleet/locomotion/tree/version_4.0.10), which probably doesn't have this code.

Are you sure this is the branch I need to merge into?

examples/client/Locomotion/src/services/geo.js
iftah-saar: There was a bug with the location services, but it was fixed, so I changed back.

examples/client/Locomotion/src/context/newRideContext/index.tsx
iftah-saar: This is also pulled from master.

package-lock.json
iftah-saar: package-locked.json is updated automatically when running npm install

examples/client/Locomotion/metro.config.js
iftah-saar: Ok. I created the ticket:
https://autofleet.atlassian.net/browse/AF-7788

package-lock.json
Davidi-74: I'm aware, this is not an NPM directory, just the root of the repo. 
The lockfile is in `./examples/client/Locmotion`

examples/client/Locomotion/src/context/futureRides/index.tsx
Davidi-74: Version `4.0.10` is already merged, the next version is `4.0.12`

package-lock.json
iftah-saar: Oops. Delete

examples/client/Locomotion/src/context/futureRides/index.tsx
iftah-saar: Ok

Davidi-74 (approved):

examples/client/Locomotion/src/popups/Selector/index.tsx
```diff
@@ -19,7 +18,12 @@ import {
 import { FlexCont } from '../../Components/Flex';
 import Button from '../../Components/Button';
 
-const Item = ({ item, onPress, selected }) => (
+interface SelectorItem {
+  value: string;
+  label: string;
+}
+
+const Item = ({ item, onPress, selected }: { item: SelectorItem; onPress: () => void; selected: boolean }) => (
```
yard2010, line 26: Nit: extract `{ item: SelectorItem; onPress: () => void; selected: boolean }` to `ItemProps`

examples/client/Locomotion/.eslintrc.json
```diff
@@ -49,7 +49,8 @@
       "no-restricted-syntax": "warn",
       "react/require-default-props": "off",
       "import/prefer-default-export": "off",
-      "react/react-in-jsx-scope": "off"
+      "react/react-in-jsx-scope": "off",
+      "react/function-component-definition": "off"
```
yard2010, line 53: Why turn this off? Did it use to break the linter on `master` as well?

yard2010 (approved):
Phenomenal job!

PR #920, opened 2026-04-26
fix: RN 0.76 bug fixes

iftah-saar:
## Summary
Bug fixes for the RN 0.76 upgrade, addressing crashes and regressions found during QA.

> **Update (review feedback):** trimmed scope per review — removed the reverse-geocode cache (perf-only, not a bug fix), removed the `@react-navigation/drawer` patch (it didn't eliminate the side-menu flicker — **AF-7975 stays open**; marker recovery after the drawer closes is handled by the react-native-maps patch), and dropped two `eslint-disable consistent-return` in RideNotes (now `return undefined`). Net ~230 fewer lines, two files removed.

## Linked Jira Tickets
- [AF-5499](https://autofleet.atlassian.net/browse/AF-5499) — Locomotion RN 0.76 [Android] | Crash on Contact us Phone/SMS
- [AF-5498](https://autofleet.atlassian.net/browse/AF-5498) — Locomotion RN 0.76 [Android] | Cannot upload profile image
- [AF-7895](https://autofleet.atlassian.net/browse/AF-7895) — Locomotion RN 0.76 [Android] | App frozen on payment method selection
- [AF-7896](https://autofleet.atlassian.net/browse/AF-7896) — Locomotion RN 0.76 | Ride history custom date filter crash/not working
- [AF-7897](https://autofleet.atlassian.net/browse/AF-7897) — Locomotion RN 0.76 [Android] | Crash on back button after viewing client message
- [AF-7898](https://autofleet.atlassian.net/browse/AF-7898) — Locomotion RN 0.76 [IOS] | Crash on address input
- [AF-7974](https://autofleet.atlassian.net/browse/AF-7974) — Locomotion RN 0.76 | Promo code addition UI issue on bottom sheet
- [AF-7975](https://autofleet.atlassian.net/browse/AF-7975) — Locomotion RN 0.76 [Android] | App flickering on side menu open action
- [AF-7996](https://autofleet.atlassian.net/browse/AF-7996) — Locomotion RN 0.76 [Android] | Crash after ride rating & tip screen
- [AF-7999](https://autofleet.atlassian.net/browse/AF-7999) — Locomotion RN 0.76 [Android] | Screen frozen on cancel ride
- [AF-8455](https://autofleet.atlassian.net/browse/AF-8455) — Locomotion RN 0.76 [Android] | Stop-point markers not visible after entering address
- [AF-8456](https://autofleet.atlassian.net/browse/AF-8456) — Locomotion RN 0.76 [Android] | Bottom-sheet shows mixed content on back from service-selection
- [AF-8461](https://autofleet.atlassian.net/browse/AF-8461) — Locomotion RN 0.76 | Keyboard hides note input when editing ride notes
- [AF-8509](https://autofleet.atlassian.net/browse/AF-8509) — Locomotion RN 0.76 [iOS] | App crash on launch (doesNotRecognizeSelector in Fabric)
- [AF-8657](https://autofleet.atlassian.net/browse/AF-8657) — Locomotion RN 0.76 [Android] | 2-3s delay + pickup reset after "Set location on map"

## Changes
- Bump `react-native-date-picker` 4.2.6 → 5.0.13 (fixes ride history date filter crash — AF-7896)
- Extract `PromoCodeButton` as standalone component from `PaymentButton` (fixes payment freeze — AF-7895)
- Update `ContactUs` component for RN 0.76 compatibility (fixes Phone/SMS crash — AF-5499)
- Update `ThumbnailPicker` for RN 0.76 compatibility (fixes profile image upload — AF-5498)
- Update `RideNotes`, `ChoosePaymentMethod`, `RangeDateTimePicker` for RN 0.76 compatibility
- Add side-by-side promo button container in payment row (AF-7974)
- Pin `react-native-svg` to 15.11.2 (15.15.4 uses `yoga::StyleSizeLength` which only exists in RN 0.78+ and breaks the Android C++ build on RN 0.76)
- Bump iOS deployment target from 13.0 → 15.6 to satisfy modern pod minimums
- Patch `stripe-react-native` `StripeSwiftInterop.h`: `STPPaymentStatus` forward declaration changed from `NSUInteger` to `NSInteger` to match the real Stripe SDK declaration (`@objc enum: Int`); newer Xcode/Swift generates a stricter forward declaration that catches the mismatch
- Clean up iOS `project.pbxproj` empty inputPaths/outputPaths arrays

## iOS Fabric crash fixes (AF-8509)
Two iOS-specific crashes on RN 0.76 with New Architecture enabled.

**Launch crash — `doesNotRecognizeSelector` in `RCTComponentData.mm`:**
Fabric's legacy interop layer wraps old `requireNativeComponent` views in `RCTLegacyViewManagerInteropComponentView`. When applying props, `RCTComponentData` called `objc_msgSend(target, setter, value)` in three places without first checking `respondsToSelector:`. The interop view doesn't implement all prop setters, so the unguarded send crashed at launch.

Fix: added `[target respondsToSelector:setter]` guards in `createEventSetter`, the `RCT_CASE` macro (both value and default branches), and `createNSInvocationSetter` in `RCTComponentData.mm`. Persisted via `patches/react-native+0.77.3.patch` (the previous `+0.76.9.patch` named a version that wasn't installed and never applied).

**"Unimplemented component: `<RNMapsMapView>`" pink error screen:**
`RCTAppDelegate.thirdPartyFabricComponents` returns an empty dictionary when `self.dependencyProvider` is nil. Without a provider, Fabric cannot discover third-party native components like `RNMapsMapView`, causing the map to render as a pink error screen.

Fix: set `self.dependencyProvider = [RCTAppDependencyProvider new]` in `AppDelegate.mm` before calling `super`.

## Android map location-pick flow fixes (AF-8657)
Three related bugs in the "Set location on map" → "Address selector" transition on Android.

**Root cause:** `changeBsPage()` in `ridePageStateContext` always calls `setIsExpanded(false)` internally. The previous fix called `setIsExpanded(true)` then `changeBsPage()` — React 18 batches both into a single render where the last `setIsExpanded` wins, leaving `isExpanded=false`.

**Bug 1 — 2-3s delay:** with `isExpanded=false`, the `useEffect([isExpanded])` that calls `bottomSheetRef.expand()` never fired on the `ADDRESS_SELECTOR` page, so the sheet stayed collapsed until something else triggered a re-render.

**Bug 2 — pickup reset:** `onSearchFocus` in `AddressSelector` checks `!isExpanded` and calls `initSps()` when true, which overwrites the map-selected pickup with the GPS position. With `isExpanded=false` after the transition, tapping the destination field triggered this path.

**Bug 3 — mini-drawer instead of full-screen:** same `isExpanded=false` state prevented the sheet from expanding to 100%.

Fix: in the `SET_LOCATION_ON_MAP` confirm handler, call `changeBsPage()` first, then `setIsExpanded(true)`. React batching leaves `isExpanded=true` (last call wins), the `useEffect` fires, and `bottomSheetRef.expand()` opens the sheet to full screen immediately.

## Map location reliability improvements
Latent causes of flaky address updates during map animations and location-pick flows.

**Programmatic animation flag (`mapAnimationState.ts`):**
`onRegionChangeComplete` fires for both user drags and code-driven `animateToRegion()` calls. When the app programmatically re-centers the map (e.g., on page transitions), the event handler treated it as a user drag, triggering reverse geocoding that silently overwrote the pickup address. Added a module-level `isProgrammaticAnimationInFlight()` flag: set before any `animateToRegion` call in `ActiveRide/index.js` and `newMap.js`, auto-cleared after the animation duration + 200ms. `onRegionChangeComplete` bails early when the flag is set.

## Android Fabric stability fixes (New Architecture)
With `newArchEnabled=true`, three Android-only freeze/crash bugs surfaced. iOS (still on Paper) was unaffected. Root causes traced to RCTModalHostView interop, unkeyed list children, and a poll loop that never stopped after ride cancellation.

- **Cancel-ride freeze** (AF-7999): replace `react-native` `Modal` with a `@gorhom/portal` overlay in `CompatModal`. `RCTModalHostView` leaks children into the host window on Fabric (RN 0.76 logs `Could not find generated setter for class com.facebook.react.views.modal.ReactModalHostManager`); on dismiss it left an orphan transparent native window swallowing all touches. The Portal renders into the existing `PortalProvider`, so no second Android window is created. `BackHandler` preserves the hardware-back behaviour
- **Cancel-ride poll loop** (AF-7999): `newRideContext.loadRide` clears ride state when the polled ride enters a `RIDE_FINAL_STATES` value, so the 4s poller stops re-firing screen handlers (and `changeBsPage`) every tick after cancellation. `CancellationReasonsPopup` captures `rideId` on first show so the popup can still send the cancellation reason after the global ride state is cleared. Misplaced `key` on `CancellationReasonCard` moved to the actual `.map()` direct child (`ClickableContainer`)
- **Tip-screen crash** (AF-7996) (`Cannot remove child at index N from parent ViewGroup`): `SelectableButton` now mounts both children-shape branches always and toggles `display: 'none'` on the unused side, so the tree shape is constant when the custom-tip button's `children` flips between a string and `null`. No visual change at any callsite
- **Select-ride freeze**: `RideButtons` renders a `<View />` placeholder in the `PassengersCounter` slot when pooling is disabled, keeping `RowContainer` child count stable when `chosenService.pooling` changes
- **Address-selector collapsing after pencil edit**: tapping the pencil on the ride-details bottom sheet opened the address selector fullscreen, then it visibly collapsed back to a 290px drawer ~1s later. Caused by a race between gorhom's snap-point-change reaction and the deferred `expand()` call: when snap points change `[373,'100%']`→`[290,'100%']`, gorhom queues an animation to `snapPoints[currentIndex]=290`; on Android Fabric the layout for the new snap points isn't calculated by the time the 100ms `setTimeout` fires, so `handleSnapToIndex` early-returns, and the queued 290 animation wins. Fix in `goBackToAddress`: call `bottomSheetRef.current.expand()` synchronously BEFORE `resetStateToAddressSelector` (while layout is still calculated on the old snap points) so gorhom sets `animatedNextPositionIndex=1`. When the snap-point reaction then fires, it sees a running animation and uses `snapPoints[1]` of the new snap points (= `'100%'`, same px value) and continues smoothly. Also adds `keyboardBehavior="extend"` + `keyboardBlurBehavior="restore"` to the shared `<BottomSheet>` so the sheet doesn't interactively re-snap to a smaller point when the Android keyboard rises with `adjustResize`
- Add missing `key` props on `.map()` returns: `Stars`, `Tips`, `RideType`, `FutureRides`, `FutureRidesView`, `MessagesList`, `AddressSelector`. Required by React; on Fabric these escalate from warnings to mount-tree corruption
- Replace empty `<></>` Fragments with `null` in `ActiveRide`, `RideCard/Map`, `CompletedRideOverviewPage` (Fabric reconciles `null` as nothing; an empty Fragment is still a tracked child)
- Remove unused `TipSectionContainer` styled component (debug code with `background: red`)

## Map markers + drawer interaction fixes (AF-8455; drawer flicker AF-7975 still open)
QA bug **AF-8455** — pick-up / drop-off / availability-car / virtual-station markers were invisible (or rendered as blank/partial squares) on Android Fabric after entering an address, and disappeared again when the side drawer opened/closed.

Two related root causes:

1. **react-native-maps 1.15.3 + Fabric**: on the new architecture the library captures a single bitmap of the marker's React children once and reuses it. With `tracksViewChanges={false}` from first mount, that snapshot happened **before** the SVG children committed a paint → blank bitmap. Compounded by `MapView.onDetachedFromWindow` overwriting `savedFeatures` even on a double-detach (so when the drawer animation briefly detaches the map, the recovery path had no features to restore), and `applyBridgedProps` re-running `initialRegion`/`initialCamera` on every re-attach.
2. **@react-navigation/drawer 6.7.2 on Fabric** _(diagnosed but **not** patched in this PR — see note below)_: the drawer toggles `accessibilityElementsHidden` / `importantForAccessibility` on its inner screen view every frame of the open/close animation. On Fabric this is a structural mutation that detaches the MapView's GLSurfaceView, and the conditional `hitSlop` (`isOpen ? undefined : swipeEdgeWidth`) recreates the PanGestureHandler instance, compounding the detach.

Fix:
- Upgrade `react-native-maps` 1.15.3 → 1.26.0 and add `patches/react-native-maps+1.26.0.patch` with four Fabric/Android fixes inside `MapView.java`:
  1. `onDetachedFromWindow` — only overwrite `savedFeatures` when `!features.isEmpty()` (prevents double-detach wiping the recovery snapshot)
  2. `applyBridgedProps` — gate `initialRegion`/`initialCamera` reapplication behind `!initialRegionSet`/`!initialCameraSet` flags (prevents drawer-animation re-attach from resetting the camera)
  3. `addFeatureAt` — lazy-recreate `attacherGroup` via `prepareAttacherView()` when null and the view is still attached (prevents markers being silently dropped after drawer close)
  4. `removeFeatureAt`, `removeCacheImageView`, `removeMapLoadingProgressBar`, `removeMapLoadingLayoutView` — null-parent guards to prevent NPEs during teardown
- _The `@react-navigation/drawer` patch was **removed** per review: it did not eliminate the side-menu flicker (**AF-7975**, still open) and cost swipe-to-close-from-anywhere + drawer accessibility. Marker recovery after the drawer closes is handled by the react-native-maps patch above (root cause 1); the residual flicker is tracked in AF-7975._
- Marker components — start with `tracksViewChanges={true}`, flip to `false` from an `onLayout` callback so the next snapshot captures real children. On Android additionally call `markerRef.current.redraw()` from `onLayout`:
  - `Marker/index.js` (StopPointMarker)
  - `PrecedingStopPointMarker/index.tsx` (converted to a stateful FC)
  - `VirtualStationMarker/index.tsx` (Callout kept as a direct Marker child)
  - `AvailabilityVehicle/index.tsx` (rotation stays as a style transform on the snapshot)
- `newMap.js` — when the side drawer transitions from `open` → `closed` on Android, bump a `polylineRedrawKey` to remount the route polyline; GoogleMap's GLSurfaceView resumes after the drawer animation but the polyline geometry isn't redrawn without a remount
- `pages/ActiveRide/styled.js` — set a neutral `#e5e3df` Android-only background on `PageContainer` to hide the pre-map paint flicker

## Address bottom-sheet back-navigation (AF-8456)
QA bug **AF-8456** — backing out of `SERVICE_ESTIMATIONS` to `ADDRESS_SELECTOR` briefly showed a "mix" of the two screens (payment-method content stretched at the wrong height) before settling. Same family as the pencil-edit fix above: when snap points change `[373,'100%']`→`[290,'100%']` in the same render that swaps the content component, gorhom on Fabric/Android keeps its current pixel height for a frame or two and renders the new content inside the old sheet.

- `pages/ActiveRide/index.js` — call `bottomSheetRef.current?.collapse()` **before** `resetStateToAddressSelector()` in both the hardware-back handler and the `backToMap()` SERVICE_ESTIMATIONS branch. `collapse()` tells gorhom "animate to snap[0]"; once snap points then change mid-flight, gorhom continues animating to the **new** snap[0] (290) for a smooth transition into `AddressSelector`
- Drive-by fix on the same focus effect: `!currentBsPage === BS_PAGES.SERVICE_ESTIMATIONS` (always `false`) → `currentBsPage !== BS_PAGES.SERVICE_ESTIMATIONS`, so `focusCurrentLocation()` runs again on focus. Added `currentBsPage` to the `useCallback` deps

## Ride-notes keyboard overlap (AF-8461)
QA bug **AF-8461** — opening the ride-notes popup on Android/iOS placed the input mid-screen, and the keyboard fully covered it.

- `popups/RideNotes/index.js` — replace the RN `Modal` with a `@gorhom/portal` top-anchored overlay (`STATUS_BAR_HEIGHT + 8`). The input is always above the keyboard, the backdrop `Pressable` dismisses the popup, and `BackHandler` preserves Android hardware-back behaviour. No second native window race
- `context/ridePageStateContext/index.tsx` — `changeBsPage` dismisses the Android keyboard when leaving `ADDRESS_SELECTOR`, preventing a leftover keyboard from interactively re-snapping the bottom sheet

## Hermes bootstrap polyfill
A cold-start crash appeared after enabling Fabric: LogBox/Fusebox warning routing references `setImmediate` before RN's polyfill bundle has loaded, throwing `setImmediate is not a function`.

- `polyfills/early-immediate-polyfill.js` — minimal `setImmediate` / `clearImmediate` polyfill using `setTimeout(fn, 0)`
- `metro.config.js` — prepend the polyfill via `serializer.getPolyfills` so it runs before the RN default polyfills (and therefore before any module-load warning routes through LogBox)

## Dependency pins & Android manifest
- Pin `react-native-gesture-handler` to `2.30.0` (2.31.x removes the abstract `getPointerEvents` method that RN 0.76 still calls — compile failure)
- Pin `react-native-draglist` `3.9.5`, `react-native-onesignal` `5.2.11`, `react-native-config` `1.5.3` to known-good versions
- `AndroidManifest.xml` — add `tools:replace="android:allowBackup"` (AppsFlyer 6.18.0 ships `allowBackup=true` in its manifest, the app sets `false` — without `tools:replace` the manifest merger fails) and `android:usesCleartextTraffic="true"`

## Misc stability
- `services/geo.js` — `getPosition` distinguishes permission-denied (`false`) from runtime error (`null`); `newRideContext.getCurrentLocation` consumes the new signal so the LOCATION_REQUEST sheet only opens on real permission denial
- `RideDrawer/AddressSelector/SearchBar.js` — when looking up the current stop point during a reorder drag, find by `id` instead of trusting the dragged item directly; null-guards stale references mid-animation
- `Components/SvgIcon/index.js` — pass `color` through to the inner SVG (needed for `currentColor` resolution on Android marker SVGs)

[AF-5499]: https://autofleet.atlassian.net/browse/AF-5499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AF-5498]: https://autofleet.atlassian.net/browse/AF-5498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AF-7895]: https://autofleet.atlassian.net/browse/AF-7895?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AF-7896]: https://autofleet.atlassian.net/browse/AF-7896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AF-7897]: https://autofleet.atlassian.net/browse/AF-7897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AF-7898]: https://autofleet.atlassian.net/browse/AF-7898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AF-7974]: https://autofleet.atlassian.net/browse/AF-7974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AF-7975]: https://autofleet.atlassian.net/browse/AF-7975?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


examples/client/Locomotion/ios/Locomotion.xcodeproj/project.pbxproj
```diff
@@ -349,14 +349,10 @@
 			inputFileListPaths = (
 				"${PODS_ROOT}/Target Support Files/Pods-Locomotion/Pods-Locomotion-frameworks-${CONFIGURATION}-input-files.xcfilelist",
 			);
-			inputPaths = (
```
yard2010, line 352: Why change this?

examples/client/Locomotion/src/pages/ActiveRide/RideDrawer/RideOptions/RideButtons/PaymentButton/index.tsx
```diff
@@ -46,7 +28,7 @@ const CardNameContainer = styled(View)`
     justify-content: flex-start;
     flex-direction: row;
     align-items: center;
-    width: ${({ fullWidth }) => (fullWidth ? '100%' : '55%')};
+    width: ${({ fullWidth }: { fullWidth?: boolean }) => (fullWidth ? '100%' : '55%')};
```
yard2010: Nit use generics to consume the types from the component
`styled(View)<{ fullWidth?: boolean }>`

examples/client/Locomotion/ios/Podfile
```diff
@@ -68,12 +68,60 @@ target 'Locomotion' do
...
+          c.build_settings['TOOLCHAINS'] = 'com.apple.dt.toolchain.XcodeDefault'
+          existing = c.build_settings['OTHER_LDFLAGS']
+          ldflags = existing.nil? || existing.to_s.empty? ? '$(inherited)' : existing
+          unless ldflags.to_s.include?('DT_TOOLCHAIN_DIR')
+            extra = '-L$(DT_TOOLCHAIN_DIR)/usr/lib/swift/$(PLATFORM_NAME)'
+            c.build_settings['OTHER_LDFLAGS'] = ldflags.is_a?(Array) ? (ldflags + [extra]) : "#{ldflags} #{extra}"
+          end
+        end
+      end
+      aggregate_target.user_project.save
```
yard2010: Let's talk about this and the new patches F2F.

yard2010 (commented):
Great job! Few comments

examples/client/Locomotion/ios/Locomotion.xcodeproj/project.pbxproj
iftah-saar: Empty inputPaths/outputPaths removed by CocoaPods on fresh pod install; existing inputFileListPaths/outputFileListPaths already cover them.

examples/client/Locomotion/ios/Podfile
iftah-saar: Ok

examples/client/Locomotion/src/pages/ActiveRide/RideDrawer/RideOptions/RideButtons/PaymentButton/index.tsx
iftah-saar: Fixed

examples/client/Locomotion/android/app/src/main/AndroidManifest.xml
```diff
@@ -14,6 +14,7 @@
       android:allowBackup="false"
       android:theme="@style/SplashTheme"
       android:usesCleartextTraffic="true"
+      tools:replace="android:allowBackup"
```
yard2010, line 17: Why replace? What dependency is using it?

examples/client/Locomotion/ios/Podfile
```diff
@@ -63,17 +63,47 @@ target 'Locomotion' do
...
+          config.build_settings['TOOLCHAINS'] = 'com.apple.dt.toolchain.XcodeDefault'
+          existing_ldflags = config.build_settings['OTHER_LDFLAGS']
+          ldflags = existing_ldflags.nil? || existing_ldflags.to_s.empty? ? '$(inherited)' : existing_ldflags
+          unless ldflags.to_s.include?('DT_TOOLCHAIN_DIR')
+            extra = '-L$(DT_TOOLCHAIN_DIR)/usr/lib/swift/$(PLATFORM_NAME)'
+            config.build_settings['OTHER_LDFLAGS'] = ldflags.is_a?(Array) ? (ldflags + [extra]) : "#{ldflags} #{extra}"
+          end
+        end
+      end
+    end
```
yard2010, line 107: Please explain shortly why these extra flags are needed, what dependency doesn't support c++17?

examples/client/Locomotion/patches/@react-navigation+drawer+6.7.2.patch
```diff
@@ -0,0 +1,185 @@
...
+-      importantForAccessibility: active ? 'auto' : 'no-hide-descendants',
++      // Patched: see drawer/views/modern/Drawer.tsx — toggling these props on
++      // Android Fabric causes the screen sibling sub-tree (containing the
++      // MapView) to be replaced, blanking the GLSurfaceView each time the
++      // drawer starts opening (this fires on progress > 0.05). The overlay is
++      // already opacity-animated and pointer-events-gated; the accessibility
++      // hide isn't critical.
+     } as const;
+   });
+
```
yard2010: Why is this patch needed? Can't we upgrade the library? This version is 3 years old

examples/client/Locomotion/patches/@stripe+stripe-react-native+0.59.2.patch
```diff
@@ -28,3 +28,16 @@ index 0000000..9f6758c
...
++++ b/node_modules/@stripe/stripe-react-native/ios/StripeSwiftInterop.h
+@@ -11,7 +11,7 @@
+ @protocol STPPaymentCardTextFieldDelegate;
+ @protocol STPCardFormViewDelegate;
+
+-typedef NS_ENUM(NSUInteger, STPPaymentStatus);
++typedef NS_ENUM(NSInteger, STPPaymentStatus);
+
+ // stripe_react_native-Swift.h also depends on these headers.
+ #import <React/RCTBridgeModule.h>
```
yard2010, line 43: Why is this needed?

examples/client/Locomotion/patches/react-native+0.76.9.patch
```diff
@@ -1,38 +1,21 @@
...
++      // Guard against Error.stack getter throwing on Hermes/New Architecture
++      // when the rejection is an Axios error whose prototype chain is broken.
 +      try {
 +        if (rejection.stack && typeof rejection.stack === 'string') {
 +          stack = rejection.stack;
 +        }
-+      } catch (_e) {
-+        // Error.stack getter may throw with an invalid receiver on Hermes
-       }
++      } catch (_) {}
```
yard2010, line 18: Please link to the related Axios issue being solved with this patch. Can't we upgrade Axios?

examples/client/Locomotion/patches/react-native-maps+1.26.0.patch
```diff
@@ -0,0 +1,108 @@
...
+         this.removeMapLoadingProgressBar();
+         if (this.mapLoadingLayout != null) {
+-            ((ViewGroup) this.mapLoadingLayout.getParent()).removeView(this.mapLoadingLayout);
++            ViewGroup parent = (ViewGroup) this.mapLoadingLayout.getParent();
++            if (parent != null) {
++                parent.removeView(this.mapLoadingLayout);
++            }
+             this.mapLoadingLayout = null;
+         }
+     }
```
yard2010, line 108: Why is this needed? Try to upgrade to latest patch version.

examples/client/Locomotion/polyfills/early-immediate-polyfill.js
```diff
@@ -0,0 +1,33 @@
...
+        fn.apply(undefined, rest);
+      }, 0);
+    };
+  }
+  if (typeof global.clearImmediate !== 'function') {
+    global.clearImmediate = function (id) {
+      clearTimeout(id);
+    };
+  }
+})();
```
yard2010, line 20: Is this a known issue with React Native/Metro? Please link to the issue for reference.

examples/client/Locomotion/src/context/newRideContext/reverseGeocodeCache.ts
```diff
@@ -0,0 +1,37 @@
...
+};
+
+export const setCached = (lat: number | string, lng: number | string, result: any): void => {
+  if (!result) return;
+  if (cache.size >= MAX_ENTRIES) {
+    const firstKey = cache.keys().next().value;
+    if (firstKey !== undefined) cache.delete(firstKey);
+  }
+  cache.set(makeKey(lat, lng), { result, t: Date.now() });
+};
```
yard2010: Why do we need to introduce a new caching mechanism in this PR? What problem does it solve?

examples/client/Locomotion/src/popups/RideNotes/index.js
```diff
@@ -24,56 +26,93 @@ export default ({
...
   useEffect(() => {
+    if (!isVisible) return;
     updateText(notes || '');
-    setTimeout(() => {
+    const t = setTimeout(() => {
       if (inputRef.current) {
         inputRef.current.focus();
       }
     }, 100);
+    // eslint-disable-next-line consistent-return
```
yard2010: Why disable

examples/client/Locomotion/src/popups/RideNotes/index.js
```diff
@@ -24,56 +26,93 @@ export default ({
...
+    return () => clearTimeout(t);
+  }, [isVisible]);
+
+  useEffect(() => {
+    if (!isVisible) return;
+    const sub = BackHandler.addEventListener('hardwareBackPress', () => {
+      onCancel();
+      return true;
+    });
+    // eslint-disable-next-line consistent-return
```
yard2010: Same

yard2010 (commented):
Few questions and suggestions

examples/client/Locomotion/android/app/src/main/AndroidManifest.xml
iftah-saar: **What AppsFlyer is**: our marketing-attribution SDK. It tells us where installs come from — which ad campaign or channel drove a download (install-conversion data) — and handles marketing / deep links so a campaign link opens the right screen. It's set up in services/appsflyer.ts and started on app launch (LocomotionRouter); we're on react-native-appsflyer 6.18.0.

**Why the manifest line**: AppsFlyer's own SDK manifest sets android:allowBackup="true", and our app sets it to false. Android's manifest merger fails when the same attribute has two different values unless we say which one wins — tools:replace="android:allowBackup" keeps our false. Not a behavior change; we still don't back up app data.

examples/client/Locomotion/ios/Podfile
iftah-saar: moving to RN 0.76 with the New Architecture made the iOS build compile a lot more C++ and Swift than before — RN 0.76 core itself requires C++17, and Fabric / interop pulls in more Swift-generated headers and C++ pods. At the same time the CI machine builds with a newer, stricter Xcode (26 / clang 21). The two together are what trip these flags; before the upgrade we didn't compile this much Swift/C++ and the toolchain was older, so none of it surfaced.

It's not that a pod "doesn't support" c++17 — it's the opposite. RN 0.76 core (folly/react) needs C++17+, and a few old pods don't declare a standard, so they default to something older and fail to compile RN headers. The loop just sets a c++17 floor. fmt is forced to c++17 because with c++20 it hits clang 21's stricter consteval errors.

examples/client/Locomotion/patches/@react-navigation+drawer+6.7.2.patch
iftah-saar: You're right to push on this — I removed it.

examples/client/Locomotion/patches/@stripe+stripe-react-native+0.59.2.patch
iftah-saar: Upgrading Stripe doesn't help — newer versions ship the same file with the same mismatch; it's a strict-compiler thing, not a Stripe bug. (The other two files in the patch are unrelated leftovers needed for older React Native.)

examples/client/Locomotion/patches/react-native+0.76.9.patch
iftah-saar: It's really a Hermes thing, not an Axios bug: Hermes' Error.stack getter throws "called with an invalid receiver" for some non-standard error objects (hermes #1496, react-native #43636). RN's promise-rejection tracker reads rejection.stack unguarded, so when one of those rejections bubbles up (Axios network errors are our most common trigger) the getter throws inside RN's own code and hides the real rejection behind a crash. The fix is a 3-line try/catch around that read — defensive in RN core, so it covers any library that hits the Hermes getter, not just Axios.

On upgrading Axios: we're on 0.27.2 which is old, but 0.x → 1.x is a breaking change across the whole app (error/response shapes) and out of scope here — and it still wouldn't fix the root cause, since the Hermes getter can throw for other error objects too. Worth a separate ticket for the Axios bump.

https://github.com/facebook/hermes/issues/1496
https://github.com/facebook/react-native/issues/43636

examples/client/Locomotion/patches/react-native-maps+1.26.0.patch
iftah-saar: The patch is four Android/Fabric-only fixes in MapView.java: don't wipe savedFeatures on a double-detach, gate initialRegion/initialCamera re-apply behind the *Set flags, lazily recreate attacherGroup so markers aren't silently dropped after the drawer closes, and null-parent guards to avoid NPEs on teardown. These are the root cause of AF-8455 (markers invisible / map blank after entering an address, and again on drawer open/close). They're not fixed upstream.

On the version: good catch — 1.27.2 exists (we're on 1.26.0). Bumping is reasonable, but it won't remove the patch (these Java fixes aren't in 1.27.x). I'd rather bump + re-roll the patch as a small follow-up than re-test the whole map flow this late in the PR — the big jump (1.15.3 → 1.26.0) is the one this PR already validated.

**react-native-maps is on the 16 KB remediation list — that's the natural place to do the version bump**

examples/client/Locomotion/polyfills/early-immediate-polyfill.js
iftah-saar: Yes, known issue — react-native #49739 "EarlyJsError: Property 'setImmediate' doesn't exist" (also hit by tamagui #3162 and wix/react-native-navigation #8082). On 0.76 + New Arch + Hermes, an early warning can route through LogBox/Fusebox before InitializeCore lazily defines setImmediate, so Hermes throws and startup bricks (red screen on Android / white on iOS).

The fix registers a tiny setTimeout-backed setImmediate/clearImmediate as the first Metro polyfill, so it exists before anything can route a warning; RN's normal lazy polyfill replaces it right after.

https://github.com/facebook/react-native/issues/49739
https://github.com/tamagui/tamagui/issues/3162
https://github.com/wix/react-native-navigation/issues/8082

examples/client/Locomotion/src/context/newRideContext/reverseGeocodeCache.ts
iftah-saar: Agreed it's out of scope here — I removed it. It was a perf-only thing (caching reverse-geocode results so a repeated pin-drag skips the network round-trip), not a crash/regression fix. The "Set location on map" bug fix (AF-8657) is the changeBsPage ordering change and doesn't depend on the cache, so nothing else moves. If we still want the latency win I'll bring it back in a separate perf PR.

examples/client/Locomotion/src/popups/RideNotes/index.js
iftah-saar: Removed the disables. Each effect had an early if (!isVisible) return; next to a cleanup return — a bare return; counts as "no value" while the cleanup returns a function, and that's what tripped consistent-return. Changing the guard to return undefined; makes both branches return a value, so the rule passes with no suppression. Same behavior.

examples/client/Locomotion/src/pages/ActiveRide/index.js
```diff
@@ -257,6 +263,13 @@ const RidePage = ({ mapSettings, navigation }) => {
     [BS_PAGES.SET_LOCATION_ON_MAP]: () => (
       <ConfirmPickup onButtonPress={(sp) => {
         updateRequestSp(sp, selectedInputIndex);
+        // changeBsPage sets isExpanded=false + correct ADDRESS_SELECTOR snap
+        // points. We immediately override isExpanded back to true so the
+        // useEffect([isExpanded]) fires and expands to full screen, and so
+        // onSearchFocus doesn't see isExpanded=false and call initSps()
+        // (which would reset the pickup we just set).
```
yard2010: Remove this comment and every other comment like this introduced in this PR

examples/client/Locomotion/metro.config.js
```diff
@@ -7,8 +7,20 @@
...
+    // Prepend our setImmediate/clearImmediate fallback before RN's default
+    // polyfill chain so it runs before any module-load warning can route into
+    // LogBox and hit the Hermes "Property 'setImmediate' doesn't exist" race
+    // on RN 0.76 + Fabric bootstrap. Once InitializeCore runs, RN's lazy
+    // polyfillGlobal definitions replace these stubs.
+    getPolyfills: () => [
+      path.resolve(__dirname, 'polyfills/early-immediate-polyfill.js'),
+      ...getRnPolyfills(),
+    ],
+  },
```
yard2010, line 19: Let's check why this is needed and remove it if it's not required.

yard2010 (approved):
Just 2 last comments

PR #934, opened 2026-06-03
feat(android): Android 16 KB page-size compliance [AF-2875]

iftah-saar:
## Summary

Brings the Locomotion Rider app into compliance with Google's 16 KB memory page-size requirement (Play policy; hard deadline Feb 1 2027). [AF-2875](https://autofleet.atlassian.net/browse/AF-2875)

Base is \`version_4.0.12_updated_iftah\` (RN 0.77.3, New Architecture ON — AF-5227, #935).

---

## 16 KB compliance — AF-2875

- **NDK r27.1 → r28** (\`android/build.gradle\`, \`28.0.12433566\`) — r28 aligns ELF LOAD segments to 16 KB by default for all CMake-built native libs. RN 0.77.3 (already in base) ships 16 KB-aligned prebuilt Hermes/RN-core \`.so\`. AGP 8.6 auto-zipaligns. No dep bumps or manifest flags needed.
- **\`lint { checkReleaseBuilds false }\`** (\`android/app/build.gradle\`) — AGP 8.6 + Stripe \`IncompatibleClassChangeError\` in \`lintVitalAnalyzeRelease\` blocks \`assembleRelease\`; required to produce the release APK the 16 KB audit runs against. \`./gradlew lint\` still works for dev.

### Audit result (arm64-v8a, release APK)
```
TOTAL: 23 ✓ / 0 ✗   (all native libs 16 KB-aligned)
zipalign -c -P 16 4: Verification successful
libjsc.so: absent ✓ (Hermes-only)
libhermes.so / libreanimated.so / librnscreens.so / libgesturehandler.so / libreactnative.so: ✓ 0x4000
```

---

## Login-screen stability (iOS New Architecture)

These fixes are kept because base already has \`newArchEnabled=true\` on both platforms and would have the same issues without them.

- **\`src/Components/PhoneNumberInput/index.tsx\` + \`src/pages/Profile/Phone.js\`** — \`autoFocus={false}\` + synchronous \`defaultCode\` seed from Config. Focusing a \`TextInput\` on mount deadlocks the JS thread on iOS Fabric, freezing the phone login screen.
- **\`src/services/MccMnc.js\`** — 3 s timeout on the native carrier-info lookups. On New Arch the native promises can hang, which blocks \`getInputIsoCode\` forever and leaves the phone input blank.
- **\`src/pages/Profile/Captcha/index.tsx\`** — 30 s watchdog + idempotent verify/close/error. The library fires \`onClose\` synchronously before \`onVerify\`; the guard ensures exactly one outcome wins and stops a hung challenge from spinning the login button forever.
- **\`src/services/network.js\`** — applies the 30 s request timeout (constructor bug meant none was ever set); spreads settings correctly.

---

## Jira
[AF-2875](https://autofleet.atlassian.net/browse/AF-2875)

[AF-2875]: https://autofleet.atlassian.net/browse/AF-2875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


examples/client/Locomotion/android/build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
         targetSdkVersion = 35
         supportLibVersion = "35.0.0"
         kotlinVersion = "2.0.21"
-        ndkVersion = "27.1.12297006"
+        ndkVersion = "28.0.12433566"
```
iftah-saar, line 11: **16 KB core (AF-2875) — NDK r27.1 → r28.** NDK r28 builds every CMake native lib with 16 KB-aligned ELF LOAD segments by default. Verified on the release APK: 23/23 arm64-v8a `.so` aligned to `0x4000`, `zipalign -c -P 16` passes, `libjsc.so` absent (Hermes).

examples/client/Locomotion/android/app/build.gradle
```diff
@@ -90,6 +90,10 @@ android {
             keyPassword properties.getProperty('keyPassword')
         }
     }
+    lint {
+        checkReleaseBuilds false
```
iftah-saar, line 94: **Lint workaround — required to produce the release build.** AGP 8.6 + the Stripe SDK throw an `IncompatibleClassChangeError` in the Lint UAST visitor during `lintVitalAnalyzeRelease`, which breaks `assembleRelease` (the build the 16 KB audit runs against). `./gradlew lint` still runs for dev. AGP/Stripe-related rather than strictly 16 KB, but needed here.

examples/client/Locomotion/metro.config.js
```diff
@@ -7,8 +7,15 @@
 
 const path = require('path');
 const { getDefaultConfig, mergeConfig } = require('@react-native/metro-config');
+const getRnPolyfills = require('@react-native/js-polyfills');
 
 const config = {
+  serializer: {
+    getPolyfills: () => [
```
iftah-saar, line 14: **setImmediate bootstrap guard (RN-0.77 New Arch).** Prepends a tiny `setImmediate`/`clearImmediate` shim (new file `polyfills/early-immediate-polyfill.js`) before RN's polyfill chain. Guards a Hermes bootstrap race where LogBox/Fusebox can touch `setImmediate` before `InitializeCore` defines it → red screen (Android) / white screen (iOS) on cold boot. Still open upstream (facebook/react-native#49739). Best removal candidate later, pending a Hermes **release** cold-boot test on 0.77.3.

examples/client/Locomotion/patches/@react-navigation+drawer+6.7.2.patch
```diff
@@ -0,0 +1,185 @@
...
+   };
+ 
+-  // FIXME: Currently hitSlop is broken when on Android when drawer is on right
+-  // https://github.com/software-mansion/react-native-gesture-handler/issues/569
+-  const hitSlop = isRight ?
+-  // Extend hitSlop to the side of the screen when drawer is closed
+-  // This lets the user drag the drawer from the side of the screen
+-  {
++  // Patched: see src/views/modern/Drawer.tsx — hitSlop kept stable to stop
++  // PanGestureHandler from reinitializing (which detaches MapView).
```
iftah-saar: **Drawer patch — stops the map blanking on drawer open/close (RN-0.77 / Fabric).** `@react-navigation/drawer` toggles `hitSlop` and `accessibilityElementsHidden`/`importantForAccessibility` on every open/close; on Android Fabric that makes RN replace the screen sub-tree, firing onDetached/onAttached on react-native-maps' `MapView` and blanking the GLSurfaceView. This keeps those props stable.

**Can't be dropped by upgrading:** drawer v7 has identical behavior (renames to `aria-hidden`, same native result) and needs a full `@react-navigation/native@7` migration; the root cause is unfixed in react-native-maps (#5557/#5743 closed *not planned*, #5907 open on 1.27.2). Trade-off: swipe-to-close now works only from the screen edge / overlay tap / hamburger.

examples/client/Locomotion/src/Components/Thumbnail/index.js
```diff
@@ -73,7 +71,7 @@ const myThumbnail = (props) => {
             }}
           />
         ) : (
-          <ImageComponent
+          <Image
```
iftah-saar: **16 KB (AF-8873) — removed `react-native-fast-image` → stock `<Image>`.** FastImage bundles a native lib and pulls SDWebImage / SDWebImageWebPCoder / libwebp into the iOS pods; swapping in RN's `<Image>` drops them. The dependency is also removed from `package.json`.

examples/client/Locomotion/src/context/newRideContext/index.tsx
```diff
@@ -703,6 +704,9 @@ const RidePageContextProvider = ({ children }: {
         };
       }
 
+      const cached = getCached(location.lat, location.lng);
```
iftah-saar: **Reverse-geocode cache.** Uses the new `reverseGeocodeCache.ts` (5-min TTL, 200-entry cap): the same coordinate dragged twice returns instantly instead of re-hitting the network, and we don't re-geocode/overwrite a pickup we already resolved. Android geocode round-trip latency is the main driver.

examples/client/Locomotion/src/pages/Profile/Captcha/index.tsx
```diff
@@ -26,44 +27,101 @@ const Captcha = ({
   const { shouldHideCaptcha, fetchHideCaptchaSetting } = useContext(OnboardingContext);
   const { user } = useContext(UserContext);
   const recaptchaRef = useRef<RecaptchaHandles | null>(null);
+  const watchdogRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const verifiedRef = useRef(false);
```
iftah-saar, line 31: **Captcha hardening (RN-0.77 login stability).** Adds a 30 s watchdog and makes the outcome idempotent. The library fires `onClose` synchronously *before* `onVerify`/`onError`, so the close is deferred one tick and `verifiedRef` guarantees exactly one of verify/close/error/timeout wins — previously a late `onClose` could override a successful verify, or a stuck challenge left the spinner hanging forever.

examples/client/Locomotion/src/services/network.js
```diff
@@ -27,12 +27,12 @@ const formatResponseLog = function ({ data = '' }) {
 class Network {
   static defaultSettings = {
     baseURL: '/',
-    timeout: 2500,
+    timeout: 30000,
```
iftah-saar, line 30: **Request timeout fix.** Applies a 30 s timeout. Previously a constructor bug built the axios client from the empty default arg, so **no** timeout was ever set — a stalled request could hang the UI (e.g. the login spinner) indefinitely.

examples/client/Locomotion/src/services/MccMnc.js
```diff
@@ -16,17 +16,25 @@ const getMobileIsoCode = async () => {
...
   const result = MccMncList.filter({ mccmnc: mccMnc });
   if (result.length > 1) {
-    const accurateResult = result.find(r => r.countryCode === (mobileIso || defaultCountryCode));
+    const accurateResult = result.find((r) => r.countryCode === (mobileIso || defaultCountryCode));
     return accurateResult?.countryCode;
   }
   return result && result[0]?.countryCode;
 };
 
+const withTimeout = (promise, ms) => {
```
iftah-saar, line 25: **Carrier-lookup timeout (login stability).** Wraps the native MCC/MNC carrier lookups in a 3 s `withTimeout` so a hanging native call can't block the phone-entry screen.

examples/client/Locomotion/src/Components/PhoneNumberInput/index.tsx
```diff
@@ -66,7 +67,7 @@ const PhoneNumberInput = ({
     <PhoneInput
       key={defaultCode}
       value={cleanNumber(value)}
-      autoFocus={autoFocus}
+      autoFocus={false}
```
iftah-saar, line 70: **autoFocus off + explicit height (iOS New Arch / Fabric).** Focusing this library's `TextInput` on mount deadlocks the JS thread on iOS Fabric and freezes the phone screen — so `autoFocus` is disabled here and at the call site (`Phone.js`). The explicit `height: 56` (just below) is also required: the lib's internal `height: '100%'` resolves to 0 against an indefinite parent under Yoga, collapsing the input.

iftah-saar (commented):
Inline notes explaining each change, grouped as **16 KB compliance (AF-2875)** and **RN-0.77 New-Architecture stability** (carried because not yet in base). Code comments were moved here per review preference. See each thread below.

examples/client/Locomotion/patches/@react-navigation+drawer+6.7.2.patch
```diff
@@ -0,0 +1,185 @@
...
+-      importantForAccessibility: active ? 'auto' : 'no-hide-descendants',
++      // Patched: see drawer/views/modern/Drawer.tsx — toggling these props on
++      // Android Fabric causes the screen sibling sub-tree (containing the
++      // MapView) to be replaced, blanking the GLSurfaceView each time the
++      // drawer starts opening (this fires on progress > 0.05). The overlay is
++      // already opacity-animated and pointer-events-gated; the accessibility
++      // hide isn't critical.
+     } as const;
+   });
+
```
yard2010: This is a huge patch. Try to upgrade the package instead, etc

examples/client/Locomotion/src/services/MccMnc.js
```diff
@@ -16,17 +16,25 @@ const getMobileIsoCode = async () => {
...
+  });
+  return Promise.race([promise, timeout]).finally(() => clearTimeout(timer));
+};
+
 export const getInputIsoCode = async () => {
   try {
     const [mmcMnc, mobileIso] = await Promise.all([
-      getMccMnc(),
-      getMobileIsoCode(),
+      withTimeout(getMccMnc(), 3000),
```
yard2010, line 36: Is this error handled somewhere?

examples/client/Locomotion/src/services/MccMnc.js
yard2010: Don't do it if it's not necessary for the app to behave 1:1 as before.

examples/client/Locomotion/src/pages/Profile/Captcha/index.tsx
```diff
@@ -26,44 +27,101 @@ const Captcha = ({
...
+      }, CAPTCHA_WATCHDOG_MS);
     } else {
       Mixpanel.setEvent('Submit phone number, without captcha , (Config.CAPTCHA_KEY is not defined)');
       onVerified();
     }
+
+    return () => {
+      clearWatchdog();
+      cancelPendingClose();
+    };
```
yard2010, line 124: Why change the captcha logic? Is this absolutely needed?

yard2010 (commented):
Great job! See my comments, we can huddle on them.

examples/client/Locomotion/patches/@react-navigation+drawer+6.7.2.patch
iftah-saar: Took it out — trimmed this PR down to just the 16kb stuff + the login fixes. I'll handle the drawer/map flicker separately.

examples/client/Locomotion/src/services/MccMnc.js
iftah-saar: Yeah, the try/catch in getInputIsoCode catches it and falls back to defaultCountryCode, so nothing escapes.

examples/client/Locomotion/src/services/MccMnc.js
iftah-saar: Kept it for now — it's 1:1 on the happy path, the timeout only kicks in if the carrier call hangs on new arch (otherwise the phone screen just sits there empty). Can drop it if you'd rather.

examples/client/Locomotion/src/pages/Profile/Captcha/index.tsx
iftah-saar: Kept this one — it's a login-stability fix, not 16kb. The recaptcha lib fires onClose right before onVerify on success, so the old code was treating a successful verify as a close. The guard makes sure only one outcome wins, and the watchdog stops a stuck challenge from spinning the login button forever. Can trim the watchdog if you think it's too much.

yard2010 (approved):

PR #938, opened 2026-06-22
RN 0.77 / New Architecture bug fixes — AF-9395, AF-9397, e2e login unblock

iftah-saar:
QA-reported bugs on the RN 0.77 + Android 16KB branch, plus the e2e login blocker — all side effects of enabling the New Architecture (Fabric).

## Fixes
- **AF-9397 (iOS) — Contact Driver / Contact Us / photo-picker crash:** `ActionSheetIOS.showActionSheetWithOptions` is a void TurboModule that terminates the app on iOS under the New Architecture. Replaced every usage (`call.tsx`, `ContactUs`, `ThumbnailPicker`) with the cross-platform `Alert.alert` already used for Android. `ActionSheetIOS` is no longer used in the app.
- **AF-9395 (Android) — duplicate map compass:** react-native-maps' Fabric layer defaulted `showsCompass` to `true` (the old bridge defaulted to `false`), so the native compass reappeared next to the custom one after a Fabric re-attach. Patched the Fabric defaults to `false` (extends the existing `react-native-maps+1.26.0.patch`); covers both MapView instances.
- **Login / e2e unblock — phone validation:** `PhoneNumberInput` reused a single libphonenumber `AsYouType` instance across keystrokes, so under the New Architecture (where input events batch onto one instance) `.input()` accumulated and corrupted `isValid()` — marking valid numbers invalid and disabling Continue (which also blocked the dev-settings number the e2e logs in with, and could block real users who paste/type fast). Now uses a fresh `AsYouType` per call.
- **dev-page:** removed redundant `autoFocus` on the Debug screen inputs.

## Verification
- AF-9395: verified at build/patch level (covers both maps).
- AF-9397: ActionSheetIOS fully removed; verify on a fresh iOS TestFlight build (login → Contact Driver / Contact Us / photo picker → no crash).
- Login/e2e: verification in progress on the Android e2e build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/client/Locomotion/patches/@gorhom+bottom-sheet+5.2.14.patch
+++ b/examples/client/Locomotion/patches/@gorhom+bottom-sheet+5.2.14.patch
@@ -1,0 +1,55 @@
+diff --git a/node_modules/@gorhom/bottom-sheet/lib/commonjs/hooks/useAnimatedLayout.js b/node_modules/@gorhom/bottom-sheet/lib/commonjs/hooks/useAnimatedLayout.js
+index 973d05c..eea2ea3 100644
+--- a/node_modules/@gorhom/bottom-sheet/lib/commonjs/hooks/useAnimatedLayout.js
++++ b/node_modules/@gorhom/bottom-sheet/lib/commonjs/hooks/useAnimatedLayout.js
+@@ -87,12 +87,12 @@ function useAnimatedLayout(containerLayoutState, topInset, bottomInset, modal, s
+   }, [state, verticalInset, modal]);
+   (0, _react.useEffect)(() => {
+     _reactNative.Dimensions.addEventListener('change', ({
+-      window
++      window: windowDimensions
+     }) => {
+       state.modify(_state => {
+         'worklet';
+ 
+-        _state.window = window;
++        _state.window = windowDimensions;
+         return _state;
+       });
+     });
+diff --git a/node_modules/@gorhom/bottom-sheet/lib/module/hooks/useAnimatedLayout.js b/node_modules/@gorhom/bottom-sheet/lib/module/hooks/useAnimatedLayout.js
+index b0eec93..9e4de19 100644
+--- a/node_modules/@gorhom/bottom-sheet/lib/module/hooks/useAnimatedLayout.js
++++ b/node_modules/@gorhom/bottom-sheet/lib/module/hooks/useAnimatedLayout.js
+@@ -83,12 +83,12 @@ export function useAnimatedLayout(containerLayoutState, topInset, bottomInset, m
+   }, [state, verticalInset, modal]);
+   useEffect(() => {
+     Dimensions.addEventListener('change', ({
+-      window
++      window: windowDimensions
+     }) => {
+       state.modify(_state => {
+         'worklet';
+ 
+-        _state.window = window;
++        _state.window = windowDimensions;
+         return _state;
+       });
+     });
+diff --git a/node_modules/@gorhom/bottom-sheet/src/hooks/useAnimatedLayout.ts b/node_modules/@gorhom/bottom-sheet/src/hooks/useAnimatedLayout.ts
+index 68c674d..c47e8ac 100644
+--- a/node_modules/@gorhom/bottom-sheet/src/hooks/useAnimatedLayout.ts
++++ b/node_modules/@gorhom/bottom-sheet/src/hooks/useAnimatedLayout.ts
+@@ -106,10 +106,10 @@ export function useAnimatedLayout(
+     [state, verticalInset, modal]
+   );
+   useEffect(() => {
+-    Dimensions.addEventListener('change', ({ window }) => {
++    Dimensions.addEventListener('change', ({ window: windowDimensions }) => {
+       state.modify(_state => {
+         'worklet';
+-        _state.window = window;
++        _state.window = windowDimensions;
+         return _state;
+       });
+     });
--- a/examples/client/Locomotion/patches/react-native-maps+1.26.0.patch
+++ b/examples/client/Locomotion/patches/react-native-maps+1.26.0.patch
@@ -1,7 +1,34 @@
+diff --git a/node_modules/react-native-maps/android/src/main/java/com/facebook/react/viewmanagers/RNMapsMapViewManagerDelegate.java b/node_modules/react-native-maps/android/src/main/java/com/facebook/react/viewmanagers/RNMapsMapViewManagerDelegate.java
+index af31722..72e84ac 100644
+--- a/node_modules/react-native-maps/android/src/main/java/com/facebook/react/viewmanagers/RNMapsMapViewManagerDelegate.java
++++ b/node_modules/react-native-maps/android/src/main/java/com/facebook/react/viewmanagers/RNMapsMapViewManagerDelegate.java
+@@ -116,7 +116,7 @@ public class RNMapsMapViewManagerDelegate<T extends View, U extends BaseViewMana
+         mViewManager.setShowsBuildings(view, value == null ? true : (boolean) value);
+         break;
+       case "showsCompass":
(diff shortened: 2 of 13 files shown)


iftah-saar:
### New commit: bottom-sheet `window` crash fix + CompatModal dialogs

**🐛 Fixed: `Uncaught Error: Property 'window' doesn't exist` on app resume**

`@gorhom/bottom-sheet@5.2.14` destructures `{ window }` from a `Dimensions` change event and uses it inside a Reanimated worklet (`useAnimatedLayout.ts`). Reanimated's compiler resolves `window` to the JS global (which doesn't exist on the worklet/UI thread) instead of the captured closure variable, crashing whenever a dimension-change fired (e.g. backgrounding then reopening the app).

- **Fix:** `patch-package` patch renaming `{ window }` → `{ window: windowDimensions }`. 5.2.14 is already the latest version (no upstream fix available), so a patch is the right approach. Patch applies automatically via the existing `postinstall` hook.

**♻️ Contact dialogs → `CompatModal`**
- `ActiveRide/call.tsx`: dialog styles extracted to `styled.ts` as styled-components (matching the rest of that file).
- `ContactUs/index.js`: replaced the `Alert` action sheet with `CompatModal` [AF-9397].

**🔑 Captcha**
- Gated on `isDebugPhoneNumber` instead of `isDevSettingOn()`.

Built & installed on QA iPhone (iOS 26.5) — build succeeded.

[AF-9397]: https://autofleet.atlassian.net/browse/AF-9397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

iftah-saar:
### New commit `def7ede1` — Contact Driver crash on the error path

Crashlytics report (build 5.0.6 / 1750) showed **Contact Driver still crashing on iOS** — but it's a *different* crash from the ActionSheetIOS one already fixed:

```
Fatal Exception: facebook::jsi::JSError
… objc_exception_rethrow
   ObjCTurboModule::performVoidMethodInvocation (RCTTurboModule.mm)
main thread: RNSScreen setViewToSnapshot → unmountChildComponentView → RCTMountingManager performTransaction
breadcrumb @ crash time: HTTP Request Error … status code 500
```

**Root cause:** when `getCallNumbers()` returns an HTTP error, the `catch` block synchronously fired a native analytics call **and** mounted the generic error popup in the same frame as the request teardown. Under the New Architecture that reentrant react-native-screens mount/unmount throws an uncaught ObjC exception → app terminates. This is the same New-Arch Fabric reentrancy family as react-native-screens [#4168](https://github.com/software-mansion/react-native-screens/issues/4168) (unfixed through 4.25.2), so a version bump wouldn't help.

**Fix:** run the error-path side effects via `InteractionManager.runAfterInteractions(...)` so the screen transition mounts in a clean frame instead of reentrantly.

Note: only reproduces when `getCallNumbers()` 500s, which can't be forced on a local dev build — validate on the App Factory build / device against the real error path.

iftah-saar:
### ✅ Contact Driver crash — root-caused & fixed (commit `594e6e86`)

Captured the real error live from the device console (release build, real error path):

```
Unhandled JS Exception: 'PortalDispatchContext' cannot be null,
please add 'PortalProvider' to the root component
→ terminating due to uncaught exception facebook::jsi::JSError
```

**Root cause:** `CompatModal` rendered through `@gorhom/portal`'s `<Portal>`. `GenericErrorPopup` (shown via `setGenericErrorPopup` on the masked-number **error path**) is rendered by `ridePageStateContext`, which sits **above** `<PortalProvider>` in `LocomotionRouter` — so its Portal had no provider in its ancestor chain and threw. iOS-only (Android uses `Alert`). The earlier ActionSheetIOS removal (bc5be249) introduced the Portal dependency; this is the latent regression.

Why the earlier attempts missed it:
- `def7ede1` (defer popup via InteractionManager) — wrong lever; the issue is missing provider, not timing.
- Local "works" — `getCallNumbers()` succeeds there, so the error popup never renders.

**Fix:** rewrote `CompatModal` on React Native's built-in `<Modal>` (transparent/fade) — no provider needed, renders above everything, fixes the error popup + contact dialog + ContactUs in one place.

**Verified on-device:** forced the error popup to render at its real tree position on a build on the QA iPhone — it now **displays ("something went wrong") instead of crashing**. Since this is a JS-level React-context error (identical debug/release), that confirms the crash is gone. Dispatching a TestFlight build for final real-API confirmation.

iftah-saar:
### Cleanup: removed ineffective `InteractionManager` defer (commit `8f5c4031`)

The `def7ede1` change (deferring the contact-error popup via `InteractionManager.runAfterInteractions`) was based on my first, incorrect diagnosis. The actual root cause was `CompatModal`'s `@gorhom/portal` dependency with no provider in its ancestor chain — fixed in `594e6e86` by switching to React Native's built-in `<Modal>`. Reverted the dead defer; the catch path is back to plain `Mixpanel.setEvent(...) + onError()`.

**Net effect of the Contact Driver work on this branch:**
- ✅ `594e6e86` — the fix: `CompatModal` → RN `<Modal>` (no PortalProvider needed). Verified on-device.
- 🧹 `8f5c4031` — removed the ineffective `InteractionManager` defer.

Confirmed working end-to-end on device: no crash, and with a driver whose number differs from the rider's, the Call/SMS dialog appears with the masked proxy number. (The earlier error modal was a backend `masked-phones` 500 when rider == driver number — separate, backend-side.)

examples/client/Locomotion/patches/@gorhom+bottom-sheet+5.2.14.patch
```diff
@@ -0,0 +1,55 @@
...
+   useEffect(() => {
+-    Dimensions.addEventListener('change', ({ window }) => {
++    Dimensions.addEventListener('change', ({ window: windowDimensions }) => {
+       state.modify(_state => {
+         'worklet';
+-        _state.window = window;
++        _state.window = windowDimensions;
+         return _state;
+       });
+     });
```
yard2010, line 55: Let's try to fix the issue without patching the bottom sheet, unless you have sources for this patch

examples/client/Locomotion/patches/react-native-maps+1.26.0.patch
```diff
@@ -94,7 +121,7 @@ diff --git a/node_modules/react-native-maps/android/src/main/java/com/rnmaps/map
              this.mapLoadingProgressBar = null;
          }
      }
-@@ -1659,7 +1693,10 @@
+@@ -1659,7 +1693,10 @@ public class MapView extends com.google.android.gms.maps.MapView implements Goog
      private void removeMapLoadingLayoutView() {
          this.removeMapLoadingProgressBar();
          if (this.mapLoadingLayout != null) {
```
yard2010, line 127: Same

yard2010 (commented):
Let's try to fix the issues without patching the libraries, let's use patches as a last resort method

examples/client/Locomotion/patches/@gorhom+bottom-sheet+5.2.14.patch
iftah-saar: Hi Yarden, fully agree patches should be a last resort 🙏 I did try to avoid both, but they're bugs inside the libraries themselves that we can't fix from our own code:

Bottom sheet (the "window" crash): it's a known library bug, reported here — https://github.com/gorhom/react-native-bottom-sheet/issues/2687. It was closed with no fix, and we're already on the latest version (5.2.14), so there's nothing to upgrade to. The patch is the same workaround others in that issue use.

Maps (duplicate compass): we already pass showsCompass={false} in our code, but with the New Architecture it gets ignored and the map's built-in default (compass on) wins. I checked the latest version (1.27.2) and it still has the same default, so upgrading wouldn't fix it.

So patches really were the last resort here. Happy to drop them the moment a fixed version is released 👍

